PR #13, opened 2024-06-12
11 poc minimizer factory

andped10:
This PR is a change to how we in `Fitter` access different minimizers as described in [ADR suggestion](https://github.com/EasyScience/EasyScience/discussions/9
)
The largest changes are in the `fitter.py`

`Fitter` now relies on a factory to produce the required minimizer.

All minimizers are a child of `MinimizerBase`.

The factory will:
- only be able to produce the minimizers for which an explicit class exists.
- the produced minimizers will all fulfill the interface described by the abstract `MinimizerBase` class.
- `Fitter` should therefore only use the minimizer methods defined in `MinimizerBase`


src/easyscience/Fitting/minimizers/factory.py
```diff
@@ -0,0 +1,30 @@
+from enum import Enum
+
+from .minimizer_bumps import Bumps
+from .minimizer_dfo import DFO
+from .minimizer_lmfit import LMFit
```
rozyczko: Any of these will throw if the given minimizer is not installed.
Originally (`__init__.py`), we would allow failures but kept going after notifying the user.
Why this (inflexible) choice?

rozyczko (approved):
Using `enum` in a factory pattern makes things simple and understandable.
We probably want to warn users if any of the "standard" minimizes are not imported, though.
I'm surprised there were no tests for that.

src/easyscience/Fitting/minimizers/factory.py
andped10: The code should now be able to run with or without all the minimizer engines.

Furthermore, the factory now explicitly list the minimizers methods from the minimizer engines that are directly supported

src/easyscience/fitting/fitter.py
```diff
@@ -1,193 +1,123 @@
...
         :param fit_function: The function to be optimized against.
-        :return: None
         """
         self._fit_object = fit_object
         self._fit_function = fit_function
-        self.__initialize()
+        self._update_minimizer(DEFAULT_MINIMIZER)
 
-    def __initialize(self):
+    def create(self, minimizer_name: str = DEFAULT_MINIMIZER) -> None:
```
damskii9992, line 65: What is the purpose of the create method? Seems to be redundant now.

src/easyscience/fitting/fitter.py
```diff
@@ -1,193 +1,123 @@
...
+    def make_model(self, pars=None) -> Callable:
+        return self._minimizer.make_model(pars)
+
+    def evaluate(self, pars=None) -> np.ndarray:
+        return self._minimizer.evaluate(pars)
+
+    def convert_to_pars_obj(self, pars) -> object:
+        return self._minimizer.convert_to_pars_obj(pars)
+
+    def initialize(self, fit_object, fit_function: Callable) -> None:
```
damskii9992, line 53: Seems to me like the initialize method is simply a combined setter method for the fit_object and the fit_function. So isn't this method also redundant?

damskii9992 (commented):
A nice step in the right direction to make EasyScience easy to actually read for developers

src/easyscience/fitting/fitter.py
andped10: Yes, you are right, but I did not dare to remove it yet since it is a public method and might be used in other codes.

I will add a code comment that I should be cleaned up.

src/easyscience/fitting/fitter.py
andped10: You are right again.  I will add a code comment and we can clean up these seemingly redundant methods at a later state.

I have also added an issue: https://github.com/EasyScience/EasyScience/issues/15

PR #18, opened 2024-06-18
Replace UUIDS with unique_names

### LINKED ISSUE #32: Replace UUIDS with string names in BORG
### Discussed in https://github.com/EasyScience/EasyScience/discussions/16

<sup>Originally posted by **damskii9992** June 17, 2024</sup>
# General
In EasyScience, we have a global object called borg, based on the singleton Borg class. This object has an associated map object based on the Graph class, which contains weak references to all created object instances which inherit from BasedBase, such as any BaseObj or BaseCollection object. One of the uses of this map, is that any object can be found no matter where it lives and no matter how deeply nested in other objects it is. 

# Current implementation
Currently, when an object is created, the borg calcuates the UUID of the object, which is a unique identifier for any python instance consisting of a 128bit string of numbers and letters. This UUID is then converted into a pure number sequence and is saved as the key of a key-value pair in a weakref dictionary on the borg with the reference to the object. To query the borg to find an object currently, the following code is used:
`borg.map.get_item_by_key(borg.map.convert_id_to_key(object))`

# Proposed implementation
Instead of using UUIDs as keys to object references, I propose to use unique strings i.e. unique "names". These unique names can be set by the user or developer, but will by default by generated according to the class name of the object and the number of those objects, such as: Parameter_0, Parameter_1 etc. With this implementation, to query the borg to find an object, the following code can be used:
`borg.map.get_item_by_key(object.unique_name)`
or if the name is known:
`borg.map.get_item_by_key("Best_Parameter")`

# Advantages with new implementation

- We don't need the borg to generate the UUIDs to then again query the borg for an object
- If the name is known, you don't even need the object to query the borg
- The list of all objects created by the call `borg.map.vertices()` will become human-readable.
- The code becomes simpler and cleaner
- A user can personally set the unique id of the most important parameters for easy access

# Disadvantage with new implementation

- Will raise errors when users or developers use the same unique name for multiple object instances</div>

damskii9992:


andped10:
It is not clear to me where and how it is ensured that a `unique_name` is unique

damskii9992:
> It is not clear to me where and how it is ensured that a `unique_name` is unique

In the Graph.py file under the add_vertex method, this method is called whenever an EasyScience object is created. It is the method which actually adds the reference to the object to the global map

src/easyscience/global_object/map.py
```diff
@@ -73,124 +68,132 @@ def is_returned(self) -> bool:
...
+            if obj_type in item.type:
+                extracted_list.append(key)
+        return extracted_list
+
+    def _get_name_iterator(self, class_name: str) -> int:
+        """Get the iterator for the name generator for a class"""
+        if class_name not in self._name_iterator_dict.keys():
+            self._name_iterator_dict[class_name] = 0
+        else:
+            self._name_iterator_dict[class_name] += 1
```
rozyczko: or just
```python
        self._name_iterator_dict.setdefault(class_name, 0)
        self._name_iterator_dict[class_name] += 1
```
avoiding conditionals

src/easyscience/global_object/map.py
```diff
@@ -73,124 +68,132 @@ def is_returned(self) -> bool:
...
         return self._nested_get("returned")
 
-    def get_item_by_key(self, item_id: int) -> object:
+    def _nested_get(self, obj_type: str) -> List[str]:
+        """Access a nested object in root by key sequence."""
+        extracted_list = []
+        for key, item in self.__type_dict.items():
+            if obj_type in item.type:
+                extracted_list.append(key)
+        return extracted_list
```
rozyczko: or just
```python
return [key for key, item in self.__type_dict.items() if obj_type in item.type]

```

src/easyscience/Objects/ObjectClasses.py
```diff
@@ -64,6 +67,21 @@ def __reduce__(self):
...
 
+    @property
+    def unique_name(self) -> str:
+        """ Get the unique name of the object."""
+        return self._unique_name
+
+    @unique_name.setter
+    def unique_name(self, new_unique_name: str):
+        """ Set a new unique name for the object. The old name is still kept in the map. 
+
```
rozyczko, line 78: so we are not able to reuse the unique names then?
why not delete (prune) the vertex? Leaving old mapping seems wasteful.

src/easyscience/Objects/Variable.py
```diff
@@ -179,6 +182,25 @@ def __reduce__(self):
...
+    @unique_name.setter
+    def unique_name(self, new_unique_name: str):
+        """ Set a new unique name for the object. The old name is still kept in the map. 
+        
+        :param new_unique_name: New unique name for the object"""
+        if not isinstance(new_unique_name, str):
+            raise TypeError("Unique name has to be a string.")
+        self._unique_name = new_unique_name
+        self._global_object.map.add_vertex(self)
+
```
rozyczko, line 203: You have 3 identical methods `unique_name` and `_unique_name_generator` in
- Variable
- ObjectClasses
- descriptor_base

Since all of them inherit from a base class, which eventually is ComponentSerializer, maybe it would make sense to combine some of these duplicates?

src/easyscience/Objects/virtual.py
```diff
@@ -164,18 +167,19 @@ def virtualizer(obj: BV) -> BV:
         d = obj.encode_data()
         if hasattr(d, 'fixed'):
             d['fixed'] = True
+        d['unique_name'] = None
         v_p = cls(**d)
```
rozyczko, line 171: `v_p` looks awfully cryptic, can't we rename this to something more meaningful?

src/easyscience/__init__.py
```diff
@@ -8,9 +8,10 @@
...
+from easyscience.global_object import GlobalObject
 
 ureg = pint.UnitRegistry()
-borg = Borg()
-borg.instantiate_stack()
-borg.stack.enabled = False
+global_object = GlobalObject()
+global_object.instantiate_stack()
+global_object.stack.enabled = False
+borg = global_object # alias for global_object, remove later
```
rozyczko: this assignment could be made a property, which returns `global_object` but also notifies the user (print()? log.warning()?) about looking obsolescence of `borg`.
Could be useful when porting the libraries

src/easyscience/global_object/hugger/__init__.py
rozyczko, line 1: I am not sure we even need anything in `hugger`.
This was going to be an impelentation of some sort of Variable wrapper so all changes to it are saved on the Undo stack.
I don't think it is fully/at all enabled so maybe we should make it go obsolescent.

rozyczko (approved):
Looks like it's ready to merge. There are minor issues to be looked at but no major problems.

src/easyscience/global_object/map.py
damskii9992: This is indeed better. Changing this.

src/easyscience/global_object/map.py
damskii9992: Yes, list comprehension is indeed better. This was not a part of my PR and the method might be completely removed or changed later on, but lets change it for now.

src/easyscience/Objects/ObjectClasses.py
damskii9992: We are, but only once the old object using the unique_name has been deleted and garbage collection has run and thus removed its entries from the map. 
Alternatively the user can manually prune the old unique_name from the map. The prune method is not "hidden".
We want to encourage users to not reuse unique_names, as that makes them not really "unique". 
But maybe this discussion belongs in the ADR Suggestion #16.

src/easyscience/Objects/Variable.py
damskii9992: It would, it would in general make a lot of sense for us to fix our inheritance tree for stuff such as this. I would claim that this does not belong in a ComponentSerializer class, as that class is just for serialization, but rather in a base class which all our objects inherit from, which we don't currently have, as our descriptors and parameters do not inherit from `BasedBase`.
So this will hopefully be amended in a future PR, when we actually fix our inheritance tree.

src/easyscience/Objects/virtual.py
damskii9992: Sure, feel free to change it in a separate branch/PR. I don't have the bigger picture of what this part of the code is actually used for. I think `v_p` stands for "virtual_parameter", possibly?

src/easyscience/__init__.py
damskii9992: Good idea.

src/easyscience/global_object/hugger/__init__.py
damskii9992: I would love to remove it, but unfortunately, it is used, specifically in the logger. Maybe we should create an issue on this to see if it can be changed in the future?

PR #19, opened 2024-06-25
New variable split into descriptor -bool, -number, -str and parameter that inherits from descriptor-number

### LINKED ISSUE #33: New Descriptors and Parameter
### Discussed in https://github.com/EasyScience/EasyScience/discussions/20

<sup>Originally posted by **andped10** June 25, 2024</sup>
# General
To build a model we need to be able to specify variables of various types. The variables used in the model can be divided into two groups. The ones that are assigned a value and remain unchanged, `Descriptors`, and the other ones that change, `Parameter`.  The latter tend to be the ones that are fitted to get a match between experimental observations and model predictions.

```mermaid
graph LR
DescriptorBase --> DescriptorNumber
DescriptorBase --> DescriptorStr
DescriptorBase --> DescriptorBool
DescriptorNumber --> Parameter
subgraph Non-fittable
   DescriptorNumber
   DescriptorStr
   DescriptorBool
end
subgraph Fittable
   Parameter
end
```

# `Descriptors`
To handle variables that can be considered "static" (rarely change after the model has been constructed) we use descriptors.  The types of descriptors in use are `bool`, `string`, or `number` values, which all share some base functionality contained in a base class. 

## `DescriptorBase`
All type specific descriptors inherit from this base class. The main properties of the base class are:
- `name`, name of this object (decorated with undo/redo functionality)
- `display_name`, optional pretty name for the object (decorated with undo/redo functionality)
- `description`, optional summary of what this object is for
- `url`, optional lookup url for documentation/information
- `parent`, optional link to the object which this descriptor is attached to

## `DescriptorBool` 
The `bool`  descriptor holds the value of a `builtin bool` type. The additional property is:
- `value`, utilized to hold the value (decorated with undo/redo functionality)

## `DescriptorStr` 
The `str` descriptor holds the value of a `builtin string` type. The additional property is:
- `value`, utilized to hold the value  (decorated with undo/redo functionality)

## `DescriptorNumber`
The `number` is different from the two former descriptors as it is continuous.  Furthermore it can be assigned a physical unit and a variance for the value.  To provide this functionally the value is kept in a [`scipp scalar`]. The additional properties are:(https://scipp.github.io/generated/functions/scipp.scalar.html)
- `value`, the numerical value of the variable (decorated with undo/redo functionality)
- `unit`, optional physical unit for the numerical value (decorated with undo/redo functionality)
- `variance`, optional variance for the value  (decorated with undo/redo functionality)
- `full_value`, the full scipp scalar holding all the above mentioned properties (decorated with undo/redo functionality)

To minimize user errors and simplify constraints, the `unit` and `full_value` cannot be changed after initialization. The unit can be converted between allowed types with the convert_unit method.

# `Parameter`
The `Parameter` class inherits from `DescriptorNumber` and serves to hold model variables that change. Such changes could be when fitting the model towards experimental data. The additional properties are:
- `min`, optional minimum for the interval for the value (decorated with undo/redo functionality)
- `max`, optional maximum for the interval for the value (decorated with undo/redo functionality)
- `free` (was `fixed`), optional sets if the parameter is allowed to vary while fitting (decorated with undo/redo functionality)
- `unconstrained` (was `enabled`), optional sets if the objects value can be set or if constraints are applied (decorated with undo/redo functionality)
- `callback`, optional callback that serves as a link between the object and a potential `Calculator`
</div>

andped10:
This PR introduce a new structure for our descriptors (bool, string, number) and parameter (number that can be fitted).

The new classes should provide all the functionality that currently is provided by `Variable.py`.  For now the new code lives alongside the existing code `Variable.py` but over time it should take over the responsibility from `Variable.py` and the old module should be removed completely.

It is a priority to make the code more readable and to include testing.

Scipp scalar is now used to hold `Number` values.

Issue with the required clean up: https://github.com/EasyScience/EasyScience/issues/23


src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -0,0 +1,428 @@
...
+            self.max = old_max
+            raise ValueError(f'Current paramter value: {self._scalar.value} must be within {new_bound=}')
+
+        # Enable the parameter if needed
+        if not self.enabled:
+            self.enabled = True
+        # Free parameter if needed
+        if self.fixed:
+            self.fixed = False
+
```
damskii9992, line 341: Do we need the bounds method when we have min/max methods?

damskii9992:
I have added input checks to inits and setters. I also added a constructor to `DescriptorNumber` to construct it from a scipp scalar.

henrikjacobsenfys (approved):

damskii9992 (approved):

src/easyscience/Objects/new_variable/parameter.py
andped10: I don't think so.  However, I would like to keep it for now to make the code as backward compatible as possible.

PR #41, opened 2024-08-01
38 add tests for minimizer bumps

andped10:
Tests for Bumps.

Changed order of method in Bumps to have the public methods appear first.  No changes to logic

Changed how imports of  `new_parameter` is done in `minimizer_lm_fit.py` to prevent circular imports when importing `Fitter`.

### DIFF
--- a/tests/unit_tests/Fitting/minimizers/test_minimizer_bumps.py
+++ b/tests/unit_tests/Fitting/minimizers/test_minimizer_bumps.py
@@ -35,7 +35,7 @@ class TestBumpsFit():
 
     def test_available_methods(self, minimizer: Bumps) -> None:
         # When Then Expect
-        assert minimizer.available_methods() == ['amoeba', 'de', 'dream', 'newton', 'scipy.leastsq', 'lm', 'pt']
+        assert minimizer.available_methods() == ['amoeba', 'de', 'dream', 'newton', 'scipy.leastsq', 'lm']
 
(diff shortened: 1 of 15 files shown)


tests/unit_tests/Fitting/minimizers/test_minimizer_bumps.py
```diff
@@ -0,0 +1,194 @@
...
+            Bumps(
+                obj='obj',
+                fit_function='fit_function', 
+                method='not_leastsq'
+            )
+
+    def test_available_methods(self, minimizer: Bumps) -> None:
+        # When Then Expect
+        assert minimizer.available_methods() == ['amoeba', 'de', 'dream', 'newton', 'scipy.leastsq', 'lm', 'pt']
+
```
rozyczko, line 39: The `pt` method in Bumps is experimental and shouldn't really be used. Also, why is `scipy` here?

andped10: OK, we will then have to filter the raw output from bumps.

I guess it is called `scipy.leastsq` because it utilizes the scipy implementation.  From the bumps code it appears that  `LevenbergMarquardtFit` is called scipy.leastsq but there is also an internal version that is referred to as `lm`.  The name for this minimizer will be left as is.

rozyczko (approved):
Looks OK. Good improvement to the code.

PR #43, opened 2024-08-01
27 arithmetic operations for parameters

### LINKED ISSUE #31: Arithmetic operations for Parameters and DescriptorNumbers
### Discussed in https://github.com/EasyScience/EasyScience/discussions/27

<sup>Originally posted by **damskii9992** July 15, 2024</sup>
# General
Adding capability to do arithmetic operations on `Parameters` and `DescriptorNumbers` allow users to use fitted parameters to do quick and easy calculations with uncertainties and units to calculate items of interest, as well as quickly and easily create new parameters/descriptors from previous ones.

# Current implementation
This is not currently implemented

# Proposed implementation
Add the following methods to `Parameter` and `DescriptorNumber`:
```python
__add__
__sub__
__mul__
__truediv__
__pow__
__neg__
__abs__
```
The output should be a `Parameter` with automatically calculated min/max if any of the elements of the arithmetic operation is a Parameter, otherwise the output should be a `DescriptorNumber`. In either case the unit and variance is also automatically propagated.</div>

damskii9992:


src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -225,3 +226,161 @@ def as_dict(self) -> Dict[str, Any]:
...
+            new_value = self.full_value  + other
+            name = self.name + ' + ' + str(other)
+        elif type(other) == DescriptorNumber:
+            original_unit = other.unit
+            try:
+                other.convert_unit(self.unit)
+            except UnitError:
+                raise UnitError(f"Values with units {self.unit} and {other.unit} cannot be added") from None
+            new_value = self.full_value + other.full_value
+            name = self._name + ' + ' + other._name
```
andped10: I think this name is a little dangerous as it could give the impression that the value of the new object would update if one of the objects, from which it was calculated, changes value.  To my understanding the value of the new object is calculated ones for all.

Maybe just create the name from the values.

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -85,6 +86,8 @@ def __init__(
             raise ValueError(f'{value=} can not be less than {min=}')
         if value > max:
             raise ValueError(f'{value=} can not be greater than {max=}')
+        if min == max:
+            raise ValueError('The min and max bounds cannot be identical. Please use fixed=True instead to fix the value.')
```
rozyczko, line 91: it should probably be `math.isclose()` or `numpy.isclose()` rather than `==`

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -246,6 +249,8 @@ def min(self, min_value: numbers.Number) -> None:
         """
         if not isinstance(min_value, numbers.Number):
             raise TypeError('`min` must be a number')
+        if min_value == self._max.value:
+            raise ValueError('The min and max bounds cannot be identical. Please use fixed=True instead to fix the value.')
```
rozyczko, line 256: cf. above

src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -193,8 +194,8 @@ def convert_unit(self, unit_str: str):
             raise TypeError(f'{unit_str=} must be a string representing a valid scipp unit')
         try:
             new_unit = sc.Unit(unit_str)
-        except Exception as message:
-            raise ValueError(message)
+        except UnitError as message:
+            raise UnitError(message) from None
```
andped10, line 202: What does the `from None` do?

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +425,298 @@ def __repr__(self) -> str:
...
+        return Parameter.from_scipp(name=name, full_value=new_value, min=min_value, max=max_value)
+
+    def __sub__(self, other: Union[DescriptorNumber, Parameter, numbers.Number]) -> Parameter:
+        if isinstance(other, numbers.Number):
+            if self.unit != 'dimensionless':
+                raise UnitError("Numbers can only be subtracted from dimensionless values")
+            new_value = self.full_value - other
+            min_value = self.min - other
+            max_value = self.max - other
+            name = f"{self.name} - {other}"
```
andped10: why new format to generate the `name` string

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +425,298 @@ def __repr__(self) -> str:
         s.append('bounds=[%s:%s]' % (repr(self.min), repr(self.max)))
         return '%s>' % ', '.join(s)
 
-    def __float__(self) -> float:
-        return float(self._scalar.value)
+    # Seems redundant
+    # def __float__(self) -> float:
+    #     return float(self._scalar.value)
+
```
rozyczko, line 434: with the assumption of  having only floats as values, this indeed is redundant.
Alternatively, we should provide other type converters.

rozyczko:
Apart from code review, a few things I found while testing

1.
`__radd__` vs. `__add__` - should be consistent
```python
    3+3+p -> "6+p"
    p+3+3 -> "p+3+3"
```

2.
Nan/inf treatment. Are the `p+np.inf` bounds correct?
```python
>>> p
<Parameter 'a': 1.0000, bounds=[-inf:inf]>
>>> p+np.inf
<Parameter 'a + inf': inf, bounds=[nan:inf]>
>>> p+np.nan
<Parameter 'a + nan': nan, bounds=[nan:nan]>
```
3.
`___div___` issue with the order (probably a scipp issue, or something wrong with `__rdiv__` vs. `__div__`)
```python
>>> a1
<Parameter 'a1': 2.0000 km, bounds=[-inf:inf]>
>>> a3
<Parameter 'a3': 2.0000 nm, bounds=[-inf:inf]>
>>> a1/a3
<Parameter 'a1 / a3': 1000000000000.0000, bounds=[-inf:inf]>
>>> a3/a1
scipp._scipp.core.UnitError: Conversion from `10e-13` to `2.18154854429043519e+244*s^(-13)*A^(-13)` is not valid.

```

src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -225,3 +226,161 @@ def as_dict(self) -> Dict[str, Any]:
...
         return raw_dict
+    
+    def __add__(self, other: Union[DescriptorNumber, numbers.Number]) -> DescriptorNumber:
+        if isinstance(other, numbers.Number):
+            if self.unit != 'dimensionless':
+                raise UnitError("Numbers can only be added to dimensionless values")
+            new_value = self.full_value  + other
+            name = self.name + ' + ' + str(other)
+        elif type(other) == DescriptorNumber:
+            original_unit = other.unit
```
andped10, line 240: I would rename to `other_unit`

src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -225,3 +226,161 @@ def as_dict(self) -> Dict[str, Any]:
         raw_dict['unit'] = str(self._scalar.unit)
         raw_dict['variance'] = self._scalar.variance
         return raw_dict
+    
+    def __add__(self, other: Union[DescriptorNumber, numbers.Number]) -> DescriptorNumber:
+        if isinstance(other, numbers.Number):
+            if self.unit != 'dimensionless':
+                raise UnitError("Numbers can only be added to dimensionless values")
+            new_value = self.full_value  + other
```
andped10, line 238: I would rename to `full_value` to make it clear that it also contains variance

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +425,298 @@ def __repr__(self) -> str:
...
+            max_value = self.max + other
+            name = f"{self.name} + {other}"
+        elif isinstance(other, DescriptorNumber):
+            original_unit = other.unit
+            try:
+                other.convert_unit(self.unit)
+            except UnitError:
+                raise UnitError(f"Values with units {self.unit} and {other.unit} cannot be added") from None
+            new_value = self.full_value + other.full_value
+            min_value = self.min + other.min if isinstance(other, Parameter) else self.min + other.value
```
andped10, line 449: Is the intended behavior that z = x + y for the limits (x_min = -10.3, x_max = -10.1) and (y_min = 10.1, y_max = 10.3) should give z (z_min = -10.3, y_max = 10.3)

henrikjacobsenfys:
```
b2=Parameter("b2", 2.0)
In [20]: b3=b2+1
In [21]: b3
Out[21]: <Parameter 'b2 + 1': 3.0000, bounds=[-inf:inf]>
In [22]: b3*2
Out[22]: <Parameter 'b2 + 1 * 2': 6.0000, bounds=[-inf:inf]>
```
Should probably be
` <Parameter '(b2 + 1) * 2': 6.0000, bounds=[-inf:inf]>`

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +425,298 @@ def __repr__(self) -> str:
...
+    
+    def __add__(self, other: Union[DescriptorNumber, Parameter, numbers.Number]) -> Parameter:
+        if isinstance(other, numbers.Number):
+            if self.unit != 'dimensionless':
+                raise UnitError("Numbers can only be added to dimensionless values")
+            new_value = self.full_value + other
+            min_value = self.min + other
+            max_value = self.max + other
+            name = f"{self.name} + {other}"
+        elif isinstance(other, DescriptorNumber):
```
andped10:
```suggestion
        elif isinstance(other, DescriptorNumber):  # Parameter inherits from DescriptorNumber and is also handled here 
```

andped10: Apply same change in other parts of the code

src/easyscience/Objects/new_variable/descriptor_number.py
andped10: Apply same change in other parts of the code

src/easyscience/Objects/new_variable/descriptor_number.py
andped10: Apply same change in other parts of the code

src/easyscience/Objects/new_variable/descriptor_number.py
andped10: Also used in other parts of the code

rozyczko:
4.
Is this as designed? 1/p -> `Parameter`, but 0/p -> `DescriptorNumber`
```python
>>> p
<Parameter 'a': 1.0000, bounds=[-inf:inf]>
>>> 0/p
<DescriptorNumber '0 / a': 0.0000>
>>> 1/p
<Parameter '1 / a': 1.0000, bounds=[-inf:inf]>
```

src/easyscience/Objects/new_variable/parameter.py
andped10: Is there a place where we can see the expected behavior of the limits ?

andped10 (commented):
Here are a first batch of suggestions / questions

src/easyscience/fitting/minimizers/minimizer_lmfit.py
```diff
@@ -18,7 +18,8 @@
 from lmfit.model import ModelResult
 
 from easyscience.Objects.new_variable import Parameter as NewParameter
-from easyscience.Objects.ObjectClasses import BaseObj
+
+# from easyscience.Objects.ObjectClasses import BaseObj ruff complained about this was unused
```
rozyczko: Cool! this fixes circular dependencies I noticed earlier and means I don't have to do a separate PR.

src/easyscience/unit_aliases.py
```diff
@@ -0,0 +1,12 @@
...
+# This document shows how to define aliases for units in scipp,
+# and how to overwrite existing aliases.
+
+sc.units.aliases.clear() # Clear existing aliases
+sc.units.aliases['funny_unit'] = sc.scalar(42.0, unit='m')
+# ^ 'funny_unit' is now an alias for 42 m and can be used as a unit in scipp and be converted to units of dimension 'm'
+
+# The representation of a unit can be changed by defining an alias for it:
+sc.units.aliases['m/ms'] = 'm/ms'
+# Setting this alias ensures that 'm/ms' is displayed as 'm/ms' instead of the default 'km/s'
```
rozyczko: This should be a part of the documentation, maybe move this blurb to README.md?

src/easyscience/Objects/new_variable/parameter.py
henrikjacobsenfys: I get the following, which I think is what it should be
```
In [60]: x=Parameter("x", -10.2, variance=2,min=-10.3,max=-10.1)

In [61]: y=Parameter("y", 10.2, variance=2,min=10.1,max=10.3)

In [62]: x+y
Out[62]: <Parameter 'x + y': 0.0000 ± 2.0000, bounds=[-0.20000000000000107:0.20000000000000107]>
```

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +425,298 @@ def __repr__(self) -> str:
...
+            return NotImplemented
+        min_value = min(combinations)
+        max_value = max(combinations)
+        parameter = Parameter.from_scipp(name=name, full_value=new_value, min=min_value, max=max_value)
+        parameter.convert_unit(parameter._base_unit())
+        return parameter
+    
+    def __truediv__(self, other: Union[DescriptorNumber, Parameter, numbers.Number]) -> Parameter:
+        if isinstance(other, numbers.Number):
+            original_other = other
```
henrikjacobsenfys: `original_other` is not needed

damskii9992:
> `__radd__` vs. `__add__` - should be consistent
> 
> ```python
>     3+3+p -> "6+p"
>     p+3+3 -> "p+3+3"
> ```
The difference here is related to how python does arithmetic operations. In the first case the interpreter first adds 3 to 3 to yield 6, and then add 6 to the parameter, using our `__radd__` with the value of 6.
In the other case we immediately use our `__add__` method to add 3 to the parameter, and then do it again, which yields a different result in the name only. This is not easy to fix, and probably not worth it, as we would have to add semi-complicated logic to how the new parameters name should be formatted.

tests/unit_tests/Objects/new_variable/test_parameter.py
```diff
@@ -1,8 +1,13 @@
 import pytest
 from unittest.mock import MagicMock
 import scipp as sc
+import numpy as np
+
+from scipp import UnitError
 
 from easyscience.Objects.new_variable.parameter import Parameter
+from easyscience.Objects.new_variable.descriptor_number import DescriptorNumber
+from easyscience.Objects.new_variable.descriptor_str import DescriptorStr
```
rozyczko: `DescriptorStr` seems unused and ruff somehow didn't catch this

damskii9992:
> Nan/inf treatment. Are the `p+np.inf` bounds correct?
> 
> ```python
> >>> p
> <Parameter 'a': 1.0000, bounds=[-inf:inf]>
> >>> p+np.inf
> <Parameter 'a + inf': inf, bounds=[nan:inf]>
> >>> p+np.nan
> <Parameter 'a + nan': nan, bounds=[nan:nan]>
> ```
I have not even considered the addition/multiplication with Inf or NaN's, as it seems like that is something the user should never be doing as it would break the minimizer anyway during fitting to have parameters with these values. Maybe we should just check this and raise an error if it is attempted? And I guess the same should be true for the setters then, as we don't want the user to be able to set a value to NaN of Inf either.

henrikjacobsenfys:
> Is this as designed? 1/p -> `Parameter`, but 0/p -> `DescriptorNumber`
> 
> ```python
> >>> p
> <Parameter 'a': 1.0000, bounds=[-inf:inf]>
> >>> 0/p
> <DescriptorNumber '0 / a': 0.0000>
> >>> 1/p
> <Parameter '1 / a': 1.0000, bounds=[-inf:inf]>
> ```

I think so, because 0/a is 0 and can not vary, so it's no longer a parameter.

rozyczko:
5.
No filtering of numerical prefixes in units

```python
>>> p1 = Parameter("p1", 2.0, unit="km*cm/J")
>>> p1
<Parameter 'p1': 2.0000 10m/N, bounds=[-inf:inf]>
>>> p1.unit
'10m/N'
```

tests/unit_tests/Objects/new_variable/test_parameter.py
andped10: Ruff is setup to ignore code in `tests`

rozyczko:
> > Is this as designed? 1/p -> `Parameter`, but 0/p -> `DescriptorNumber`
> > ```python
> > >>> p
> > <Parameter 'a': 1.0000, bounds=[-inf:inf]>
> > >>> 0/p
> > <DescriptorNumber '0 / a': 0.0000>
> > >>> 1/p
> > <Parameter '1 / a': 1.0000, bounds=[-inf:inf]>
> > ```
> 
> I think so, because 0/a is 0 and can not vary, so it's no longer a parameter.

Yup, and there even is an associated unit test `test_zero_value_dividided_by_parameter`.
Good. :thumbsup:

rozyczko:
> > Nan/inf treatment. Are the `p+np.inf` bounds correct?
> > ```python
> > >>> p
> > <Parameter 'a': 1.0000, bounds=[-inf:inf]>
> > >>> p+np.inf
> > <Parameter 'a + inf': inf, bounds=[nan:inf]>
> > >>> p+np.nan
> > <Parameter 'a + nan': nan, bounds=[nan:nan]>
> > ```
> 
> I have not even considered the addition/multiplication with Inf or NaN's, as it seems like that is something the user should never be doing as it would break the minimizer anyway during fitting to have parameters with these values. Maybe we should just check this and raise an error if it is attempted? And I guess the same should be true for the setters then, as we don't want the user to be able to set a value to NaN of Inf either.

If we screen NaNs/Infs out of our datasets, one would feel tempted to implement a masking algorithm, where masked points would be set to NaN in order to have them filtered out for fitting. This means we should allow setting non-numbers but act accordingly. Bad idea, since we normally use a second, boolean, array with masked points/pixels but this could be a possibility?

henrikjacobsenfys:
> ```python
> p1 = Parameter("p1", 2.0, unit="km*cm/J")
> ```

Just mentioned it to Christian, the solution is probably that the unit conversion should happen in the setter. Interestingly, multiplying p by 1 changes the unit:

```
In [73]: p1 = Parameter("p1", 2.0, unit="km*cm/J")

In [74]: p1
Out[74]: <Parameter 'p1': 2.0000 10m/N, bounds=[-inf:inf]>

In [75]: p1*1
Out[75]: <Parameter 'p1 * 1': 20.0000 s^2/kg, bounds=[-inf:inf]>
```

henrikjacobsenfys:
> > > Nan/inf treatment. Are the `p+np.inf` bounds correct?
> > > ```python
> > > >>> p
> > > <Parameter 'a': 1.0000, bounds=[-inf:inf]>
> > > >>> p+np.inf
> > > <Parameter 'a + inf': inf, bounds=[nan:inf]>
> > > >>> p+np.nan
> > > <Parameter 'a + nan': nan, bounds=[nan:nan]>
> > > ```
> > 
> > 
> > I have not even considered the addition/multiplication with Inf or NaN's, as it seems like that is something the user should never be doing as it would break the minimizer anyway during fitting to have parameters with these values. Maybe we should just check this and raise an error if it is attempted? And I guess the same should be true for the setters then, as we don't want the user to be able to set a value to NaN of Inf either.
> 
> If we screen NaNs/Infs out of our datasets, one would feel tempted to implement a masking algorithm, where masked points would be set to NaN in order to have them filtered out for fitting. This means we should allow setting non-numbers but act accordingly. Bad idea, since we normally use a second, boolean, array with masked points/pixels but this could be a possibility?

I can't think of a situation where anyone would want a `Parameter` to be NaN of Inf - after all, the idea of a `Parameter` is that it should be fitable, right?

damskii9992:
> `___div___` issue with the order (probably a scipp issue, or something wrong with `__rdiv__` vs. `__div__`)
> 
> ```python
> >>> a1
> <Parameter 'a1': 2.0000 km, bounds=[-inf:inf]>
> >>> a3
> <Parameter 'a3': 2.0000 nm, bounds=[-inf:inf]>
> >>> a1/a3
> <Parameter 'a1 / a3': 1000000000000.0000, bounds=[-inf:inf]>
> >>> a3/a1
> scipp._scipp.core.UnitError: Conversion from `10e-13` to `2.18154854429043519e+244*s^(-13)*A^(-13)` is not valid.
> ```
Should've fixed this issue just now. Have added logic to check if the unit prefix contains an exponential, which was why this failed.

src/easyscience/Objects/new_variable/descriptor_number.py
henrikjacobsenfys: It seems that it suppresses the original error message. Whether or not we want this is up for discussion. Let's keep it as is for now.

src/easyscience/Objects/new_variable/descriptor_number.py
henrikjacobsenfys: Done

src/easyscience/Objects/new_variable/descriptor_number.py
henrikjacobsenfys: changed to new_full_value

src/easyscience/Objects/new_variable/parameter.py
henrikjacobsenfys: Done

henrikjacobsenfys:
> 10m



> > ```python
> > p1 = Parameter("p1", 2.0, unit="km*cm/J")
> > ```
> 
> Just mentioned it to Christian, the solution is probably that the unit conversion should happen in the setter. Interestingly, multiplying p by 1 changes the unit:
> 
> ```
> In [73]: p1 = Parameter("p1", 2.0, unit="km*cm/J")
> 
> In [74]: p1
> Out[74]: <Parameter 'p1': 2.0000 10m/N, bounds=[-inf:inf]>
> 
> In [75]: p1*1
> Out[75]: <Parameter 'p1 * 1': 20.0000 s^2/kg, bounds=[-inf:inf]>
> ```

This should be fixed now

src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -225,3 +231,164 @@ def as_dict(self) -> Dict[str, Any]:
...
+        if isinstance(other, numbers.Number):
+            exponent = other
+            name = f"{self.name} ** {other}"
+        elif type(other) is DescriptorNumber:
+            if other.unit != 'dimensionless':
+                raise UnitError("Exponents must be dimensionless")
+            if other.variance is not None:
+                raise ValueError("Exponents must not have variance")
+            exponent = other.value
+            name = self.name+" ** "+other.name
```
andped10:
```suggestion
            name = self.name + " ** " + other.name
```

src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -225,3 +231,164 @@ def as_dict(self) -> Dict[str, Any]:
...
+            if self.variance is not None:
+                raise ValueError("Exponents must not have variance")
+            new_value = other ** self.value
+        else:
+            return NotImplemented
+        return new_value
+
+    def __neg__(self) -> DescriptorNumber:
+        new_value = -self.full_value
+        name = '-'+self.name
```
andped10:
```suggestion
        name = '-' + self.name
```

src/easyscience/Objects/new_variable/descriptor_number.py
```diff
@@ -225,3 +231,164 @@ def as_dict(self) -> Dict[str, Any]:
...
+        return DescriptorNumber.from_scipp(name=name, full_value=new_value)
+    
+    def __abs__(self) -> DescriptorNumber:
+        new_value = abs(self.full_value)
+        name = 'abs('+self.name+')'
```
andped10:
```suggestion
        name = 'abs(' + self.name + ')'
```

tests/unit_tests/Objects/new_variable/test_descriptor_number.py
```diff
@@ -1,6 +1,9 @@
 import pytest
 from unittest.mock import MagicMock
 import scipp as sc
+from scipp import UnitError
+
+from scipp import UnitError
```
andped10: delete

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -117,6 +122,10 @@ def __init__(
         }
         self._constraints = Constraints(builtin=builtin_constraint, user={}, virtual={})
 
+        if self.unit is not None:
```
andped10: Is needed ?  
Is it not enough `self.convert_unit` is called in the super constructor

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -85,8 +86,14 @@ def __init__(
             raise ValueError(f'{value=} can not be less than {min=}')
         if value > max:
             raise ValueError(f'{value=} can not be greater than {max=}')
+
+        if np.isclose(min, max, rtol=1e-9, atol=0.0):
+            raise ValueError('The min and max bounds cannot be identical. Please use fixed=True instead to fix the value.')
         if not isinstance(fixed, bool):
             raise TypeError('`fixed` must be either True or False')
+
```
andped10, line 94:
```suggestion
        # Must be set before calling super constructor in which `self.convert_unit` is called
```

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -246,6 +255,9 @@ def min(self, min_value: numbers.Number) -> None:
         """
         if not isinstance(min_value, numbers.Number):
             raise TypeError('`min` must be a number')
+        # if min_value == self._max.value:
```
andped10: delete

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -272,6 +284,9 @@ def max(self, max_value: numbers.Number) -> None:
         """
         if not isinstance(max_value, numbers.Number):
             raise TypeError('`max` must be a number')
+        # if max_value == self._min.value:
```
andped10: delete

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +433,299 @@ def __repr__(self) -> str:
...
+            except UnitError:
+                raise UnitError(f"Values with units {self.unit} and {other.unit} cannot be subtracted") from None
+            new_full_value = self.full_value - other.full_value
+            if isinstance(other, Parameter):
+                min_value = self.min - other.max if other.max != np.inf else -np.inf
+                max_value = self.max - other.min if other.min != -np.inf else np.inf
+            else:
+                min_value = self.min - other.value
+                max_value = self.max - other.value
+            name = self.name+" - "+other.name
```
andped10: For consistency I would do:
name = f"{self.name} - {other}"

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +433,299 @@ def __repr__(self) -> str:
...
+        if isinstance(other, numbers.Number):
+            new_full_value = other / self.full_value
+            other_value = other
+            name = f"{other} / {self.name}"
+            if other_value == 0:
+                return DescriptorNumber.from_scipp(name=name, full_value=new_full_value)
+        elif isinstance(other, DescriptorNumber): # Parameter inherits from DescriptorNumber and is also handled here
+            new_full_value = other.full_value / self.full_value
+            other_value = other.value
+            name = other.name+" / "+self.name
```
andped10: For consistency:
name = f"{other.name} / {self.name}"

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -418,5 +433,299 @@ def __repr__(self) -> str:
...
+    
+    def __truediv__(self, other: Union[DescriptorNumber, Parameter, numbers.Number]) -> Parameter:
+        if isinstance(other, numbers.Number):
+            if other == 0:
+                raise ZeroDivisionError("Cannot divide by zero")
+            new_full_value = self.full_value / other
+            combinations = [self.min / other, self.max / other]
+            name = f"{self.name} / {other}"
+        elif isinstance(other, DescriptorNumber):  # Parameter inherits from DescriptorNumber and is also handled here
+            original_value = other.value
```
andped10: I guess this should be a variable called:
other_value

tests/unit_tests/Objects/new_variable/test_parameter.py
```diff
@@ -334,4 +339,549 @@ def test_as_data_dict(self, clear, parameter: Parameter):
...
+            assert result_reverse.min == expected_reverse.min
+            assert result_reverse.max == expected_reverse.max
+
+    @pytest.mark.parametrize("test, expected, expected_reverse", [
+        (DescriptorNumber(name="test", value=2, variance=0.1, unit="cm"), Parameter("name * test", 2, "dm^2", 0.14, 0, 20), Parameter("test * name", 2, "dm^2", 0.14, 0, 20)),
+        (DescriptorNumber(name="test", value=0, variance=0.1, unit="cm"), DescriptorNumber("name * test", 0, "dm^2", 0.1), DescriptorNumber("test * name", 0, "dm^2", 0.1))],
+        ids=["regular", "zero_value"])
+    def test_multiplication_with_descriptor_number(self, parameter : Parameter, test, expected, expected_reverse):
+        # When 
+        parameter._callback = property()
```
andped10, line 609: What is this doing

andped10: same goes for the tests below

andped10 (approved):
Cool

src/easyscience/Objects/new_variable/descriptor_number.py
rozyczko: We use the `name` attribute only for GUI representation of a parameter/descriptor. If we ever are in a situation where we need to display arithmetically modified parameters, then this is a potential problem (maybe in some form of an advanced constraint editor?)
Maybe we should be using `.unique_name` as the name? Or, as Andreas suggests - just values.

tests/unit_tests/Objects/new_variable/test_parameter.py
henrikjacobsenfys: No idea! If anyone knows, I'd like to learn. @damskii9992  ?

damskii9992: You mean the parameter._callback = property()? It is to use the parameter fixture where the callback is mocked away. So here I just set it to the normal property() value to make it a normal parameter.

src/easyscience/Objects/new_variable/descriptor_number.py
damskii9992: It suppresses the full error path. I found the raised error to be much more readable without it.

src/easyscience/Objects/new_variable/parameter.py
damskii9992: for consistency

PR #47, opened 2024-08-08
slightly reworked API for accessing info about minimizers

rozyczko:
- Renamed `available_methods` to `all_methods`.
- Added `supported_methods`.
- Made both methods static and required in derived classes.
- Updated tests.


andped10 (approved):
Looks good to me

PR #49, opened 2024-08-13
48 Re-increment unique names if already exists

### LINKED ISSUE #48: Load Parameters before creating new parameters cause unique_name clashes
When starting out a project by loading previously defined parameters from an earlier session, the `global_object` map gets filled with these unique names and when the user then starts creating new parameters, the `unique_name_generator` increments the name iterators and eventually causes `unique_name` crashes with the unique names populated at load time.

This can be solved by incrementing iterators again if the `unique_name` already exists in the `global_object` when generating the unique names.

damskii9992:
Fixes an issue with the unique_name generation

### DIFF
--- a/src/easyscience/Objects/ObjectClasses.py
+++ b/src/easyscience/Objects/ObjectClasses.py
@@ -212,7 +212,11 @@ class BasedBase(ComponentSerializer):
         """
         class_name = self.__class__.__name__
         iterator_string = str(self._global_object.map._get_name_iterator(class_name))
-        return class_name + "_" + iterator_string
+        name = class_name + "_" + iterator_string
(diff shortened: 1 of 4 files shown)


andped10 (approved):
Looks good to me.

We are going to decommission `ObjectClasses` and `Variable` in foreseeable future, right?  If that is not the case additional tests need to be added.

damskii9992:
> Looks good to me.
> 
> We are going to decommission `ObjectClasses` and `Variable` in foreseeable future, right? If that is not the case additional tests need to be added.

`Variable` should be decommissioned yeah and `ObjectClasses` should be changed/re-arranged too so that the `unique_name_generator` is only defined in 1 place.

PR #61, opened 2024-08-23
60 add support lmfit differential evolution

### LINKED ISSUE #60: Add support LMFit differential_evolution
The LMFit differential_evolution is needed in the tutorials for EasyReflectometry

andped10:
The  LMFit differential_evolution method is needed in some tutorials for EasyReflectometry

### DIFF
--- a/src/easyscience/fitting/available_minimizers.py
+++ b/src/easyscience/fitting/available_minimizers.py
@@ -1,0 +1,81 @@
+import warnings
+from enum import Enum
+from enum import auto
+
+import pkg_resources
+
+installed_packages = {pkg.key for pkg in pkg_resources.working_set}
+
+lmfit_engine_available = False
+if 'lmfit' in installed_packages:
+    lmfit_engine_available = True
+else:
+    # TODO make this a proper message (use logging?)
+    warnings.warn('LMFit minimization is not available. Probably lmfit has not been installed.', ImportWarning, stacklevel=2)
+
+bumps_engine_available = False
+if 'bumps' in installed_packages:
+    bumps_engine_available = True
+else:
+    # TODO make this a proper message (use logging?)
+    warnings.warn('Bumps minimization is not available. Probably bumps has not been installed.', ImportWarning, stacklevel=2)
+
+dfo_engine_available = False
+if 'dfo-ls' in installed_packages:
+    dfo_engine_available = True
+else:
+    # TODO make this a proper message (use logging?)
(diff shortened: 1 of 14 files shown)


src/easyscience/fitting/available_minimizers.py
```diff
@@ -0,0 +1,81 @@
+import warnings
+from enum import Enum
+from enum import auto
+
+import pkg_resources
+
+installed_packages = {pkg.key for pkg in pkg_resources.working_set}
```
rozyczko, line 7: `pkg_resources` seenms to be deprecated.
```
>>> import pkg_resources
<stdin>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

Consider using
```python
import importlib.metadata
installed_packages = [x.name for x in importlib.metadata.distributions()]
```

src/easyscience/fitting/minimizers/minimizer_base.py
```diff
@@ -122,6 +122,21 @@ def evaluate(self, x: np.ndarray, minimizer_parameters: dict[str, float] = None,
...
+            method['method'] = self._method
+
+        # Set / overwrite if method was passed
+        if passed_method is not None:
+            if passed_method in self.supported_methods():
+                method['method'] = passed_method
+            else:
+                raise FitError(f'Method {passed_method} not available in {self.__class__}')
+        return method
+
```
rozyczko, line 135: How about removing `else` and streamlining this a bit with premature returns?

```python
def _get_method_dict(self, passed_method: Optional[str] = None) -> dict[str, str]:
    if passed_method is not None:
        if passed_method not in self.supported_methods():
            raise FitError(f'Method {passed_method} not available in {self.__class__}')
        return {'method': passed_method}

    if self._method is not None:
        return {'method': self._method}

    return {}

```

rozyczko (approved):
minor issues only.

src/easyscience/fitting/minimizers/minimizer_base.py
andped10: Thanks, done

src/easyscience/fitting/available_minimizers.py
andped10: thanks, done

andped10: Reverted as `importlib.metadata.distributions` is not supported in 3.9

PR #67, opened 2024-08-27
66 constraints do not seem to work when there is a calculator set

### LINKED ISSUE #66: Constraints do not seem to work when there is a calculator set
With the new parameter there is an issue when applying constraints.  The bug seems to appear only when the callback functionality is used.

andped10:
Prevent the usage of eventual callback values when applying constraints


rozyczko (approved):
This looks sensible.

One minor typo: in the definition of `def value_no_call_back` in `parameters.py`
the docstring should say `suppressing` instead of `surpassing`.

PR #71, opened 2024-09-09
Build documentation 

andped10:
The theme we use in Sphinx (`sphinx-rtd-theme`) needs to be a relative recent.  This breaks the compatibility with the installed version of `docutils`.  

When doing a forced upgrade the documentation builds and it look alright.  My suggestion is therefore that we just ignore the issue as it will probably disappear when we become able to use a more up to date version of  `docutils`

Different syntax adaptation to the new code structure.


rozyczko (approved):
Looks good.
At some point we need to revisit the change and make sure docutils/spinx themes install normally, without us forcing the upgrade.

PR #75, opened 2024-09-13
Adjustments to make the EasyReflectometryApp execute

andped10:
Cleaned up `AvailableMinimizers`

Various places we also accept `DescriptorBase` as valid type for moving on.

All classes now accepts `unique_name` as argument to constructor

Made a `__copy__` method for `ObjectClasses`

Got rid of `@id`

Moved `Constraints` to repo root to break circular dependency

Cleaned `map` class

Tests of utils are now using the new `Parameter` and `DecriptorNumber`


src/easyscience/Utils/classTools.py
```diff
@@ -59,14 +59,13 @@ def removeProp(inst: BV, name: str) -> None:
...
     start_idx = 0 + int(skip_first)
-    ids = []
+    unique_names = []
     names = []
-    model_id = model_obj.unique_name
     for par in pars:
-        elem = par.unique_name
-        route = global_object.map.reverse_route(elem, model_id)
+        par_unique_name = par.unique_name
+        route = global_object.map.reverse_route(par_unique_name, model_obj.unique_name)
```
henrikjacobsenfys: Small thing, but it seems inconsistent to give `par_unique_name` and `model_obj.unique_name` as input instead of either
`(par.unique_name, model_obj.unique_name)` or defining e.g. 
`model_obj_unique_name=model_obj.unique_name` and giving that as input

src/easyscience/fitting/available_minimizers.py
```diff
@@ -1,55 +1,65 @@
 import warnings
+from dataclasses import dataclass
 from enum import Enum
-from enum import auto
 
 import pkg_resources
 
-installed_packages = {pkg.key for pkg in pkg_resources.working_set}
+installed_packages = {pkg.key.lower() for pkg in pkg_resources.working_set}
```
henrikjacobsenfys: I don't think installed_packages is used anywhere

src/easyscience/Utils/classTools.py
andped10: Thanks, changed to just use the attribute

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -255,9 +251,7 @@ def _dfo_fit(pars: Dict[str, Parameter], model: Callable, **kwargs):
         """
 
         ## TODO clean when full move to new_variable
```
henrikjacobsenfys, line 253: I suppose this TODO can be deleted?

src/easyscience/fitting/available_minimizers.py
andped10: Thanks, left over from old way of determining if a given module with minimizers is available

src/easyscience/fitting/minimizers/minimizer_lmfit.py
```diff
@@ -164,9 +165,7 @@ def convert_to_par_object(parameter: Parameter) -> LMParameter:
         :rtype: LMParameter
         """
         ## TODO clean when full move to
```
henrikjacobsenfys, line 167: Same here

src/easyscience/fitting/minimizers/minimizer_dfo.py
andped10: This we need to keep.  The if / else statement should be deleted when fully moving to the new variable

src/easyscience/fitting/minimizers/minimizer_lmfit.py
andped10: This we need to keep. The if / else statement should be deleted when fully moving to the new variable

src/easyscience/Objects/new_variable/descriptor_base.py
```diff
@@ -27,7 +27,7 @@ class DescriptorBase(ComponentSerializer, metaclass=abc.ABCMeta):
     # Used by serializer
     _REDIRECT = {'parent': None}
 
-    _global_object = global_object
+    #    _global_object = global_object
```
henrikjacobsenfys:
```suggestion
```

henrikjacobsenfys (commented):
Looks good!

henrikjacobsenfys (approved):

PR #78, opened 2024-09-26
Upgrade numpy and pint

henrikjacobsenfys:
Removed pin to numpy and pint, and updated `__repr__` in Parameter to handle new numpy behaviour.


henrikjacobsenfys:
Warning: this drops support for Python 3.9, since it doesn't play well with scipp. So we should probably wait a while to merge this.

PR #80, opened 2024-09-27
unique name when created from dict

andped10:
We need to keep the `unique_name` in order to be able to reproduce the object.

We are still able to get a copy by using a dict where we skip the `unique_name`


henrikjacobsenfys (approved):
Looks good to me

andped10:
@damskii9992 Please confirm that these merged changes are OK

damskii9992:
> @damskii9992 Please confirm that these merged changes are OK

Changes are OK for now but might lead to issues in the future.
Explaining here for documentation: If a user has already created an object with a unique_name: "BaseObj_0" and they then try to load an object which was saved with the same unique_name, this will fail. 
To circumvent this, we might want to add an option to overwrite the loaded objects unique_name and add a proper error message to the user telling them to do this in case of name clashes upon loading.

PR #82, opened 2024-10-29
Add the property `free` to the class `Parameter` as an alias for `not fixed`

AndrewSazonov:
This is the first step in implementing the property `free` for the class `Parameter`, in line with our discussion  https://github.com/EasyScience/EasyScience/discussions/20 as well as the issue https://github.com/EasyScience/EasyScience/issues/24. 

For now, the property `free` has been added as an alias to `not fixed`.

### DIFF
--- a/src/easyscience/Objects/Variable.py
+++ b/src/easyscience/Objects/Variable.py
@@ -503,7 +503,6 @@ class Parameter(Descriptor):
         min: Optional[numbers.Number] = -np.inf,
         max: Optional[numbers.Number] = np.inf,
         fixed: Optional[bool] = False,
-        free: Optional[bool] = True,
         **kwargs,
     ):
         """
--- a/src/easyscience/Objects/Variable.py
+++ b/src/easyscience/Objects/Variable.py
@@ -503,6 +503,7 @@ class Parameter(Descriptor):
         min: Optional[numbers.Number] = -np.inf,
         max: Optional[numbers.Number] = np.inf,
         fixed: Optional[bool] = False,
+        free: Optional[bool] = True,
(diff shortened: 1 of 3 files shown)


src/easyscience/Objects/Variable.py
```diff
@@ -503,6 +503,7 @@ def __init__(
         min: Optional[numbers.Number] = -np.inf,
         max: Optional[numbers.Number] = np.inf,
         fixed: Optional[bool] = False,
+        free: Optional[bool] = True,
```
andped10: To me it seems like this is not used and I would delete it.  Also, it is probably dangerous both to `fixed` and `free` since they are mutual exclusive

src/easyscience/Objects/new_variable/parameter.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         min: Optional[numbers.Number] = -np.inf,
         max: Optional[numbers.Number] = np.inf,
         fixed: Optional[bool] = False,
+        free: Optional[bool] = True,
```
andped10: To me it seems like this is not used and I would delete it. Also, it is probably dangerous both to fixed and free since they are mutual exclusive

andped10 (approved):

src/easyscience/Objects/Variable.py
AndrewSazonov: Deleted

src/easyscience/Objects/new_variable/parameter.py
AndrewSazonov: Deleted

PR #84, opened 2024-10-30
To increase usability of minimizer it should be possible to set the tolerance and the max number of iteration

andped10:
On the Fitter object it is now possible to set 
- the maximal evaluations (LMFit and DFO function evaluations  / Bumps steps)
- the tolerance (LMFit and Bumps change in function value / Bumps change in new parameters / DFO: size of trust region)


src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -107,10 +109,16 @@ def fit(
 
         if minimizer_kwargs is None:
             minimizer_kwargs = {}
-        # else:
-        #     minimizer_kwargs = {"fit_kws": minimizer_kwargs}
         minimizer_kwargs.update(engine_kwargs)
 
+        if tolerance is not None:
+            if 0.1 < tolerance:
```
AndrewSazonov: Why is 0.1 chosen as the maximum possible tolerance value?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -261,13 +265,23 @@ def _dfo_fit(pars: Dict[str, Parameter], model: Callable, **kwargs):
...
             raise FitError(f'Fit failed with message: {results.msg}')
 
         return results
+
+    @staticmethod
+    def _prepare_kwargs(tolerance: Optional[float] = None, max_evaluations: Optional[int] = None, **kwargs) -> dict[str:str]:
+        if max_evaluations is not None:
+            kwargs['maxfun'] = max_evaluations  # max number of function evaluations
+        if tolerance is not None:
+            if 0.1 < tolerance:
```
AndrewSazonov: Why is 0.1 chosen as the maximum possible tolerance value?

AndrewSazonov (approved):
Looks good to me

PR #85, opened 2024-11-11
Allow parameter value to be changed even if parameter is not enabled

AndrewSazonov:
We need to be able to set a new value for a parameter, even if that parameter is not currently enabled.


rozyczko (approved):
The idea is to allow all parameters to be user changeable.
This probably makes sense as one should be able to modify Parameter values and not have them displayed in the list of fittables.
But maybe then `enabled` is not the right word. Maybe `fittable` would be better?
This, however, is for future ADRs and for more discussion.
As is, the code change is minimal and can be pushed.

Issue #88, opened 2024-11-13
Invalid escape sequence in docstring


jl-wynen:
There are invalid escape sequences in a docstring here: https://github.com/EasyScience/EasyScience/blob/a6d93795c396eb2c843d162908d636cbfe4e8a36/src/easyscience/Datasets/xarray.py#L284 Namely `\|`.

To fix it, either use `\\|` or make the docstring a raw string literal.

Normally, this only raises a `DeprecationWarning` which often does not show up in the terminal. But in our tests, we treat warnings as errors. This makes Python raise `SyntaxError` here.

AndrewSazonov:
Fixed in #90

PR #89, opened 2024-11-20
Fix invalid escape sequence in docstring

AndrewSazonov:


rozyczko (commented):
Isn't the double backslash redundant? Python will convert it to the single slash before even Pandoc sees it. Also, the rendering already looks correct in the docs, it seems.
I must be missing something...

AndrewSazonov:
> Isn't the double backslash redundant? Python will convert it to the single slash before even Pandoc sees it. Also, the rendering already looks correct in the docs, it seems. I must be missing something...

This is to fix #88.

rozyczko (approved):

PR #93, opened 2024-12-17
Numpy2

henrikjacobsenfys:
Updated to use the newest versions of numpy and pint. Dropped Python 3.9, and included 3.13.

I had a lot of trouble with this part of parameter.py, on line 163

```
    @property
    def value(self) -> numbers.Number:
        """
        Get the value of self as a Number.

        :return: Value of self without unit.
        """
        if self._callback.fget is not None:
            existing_value = self._callback.fget()
            if existing_value != self._scalar.value:
                self._scalar.value = existing_value
        return self._scalar.value
```
		

The problem was that the mock did not have a `_callback.fget()` function. In this case, it returns something else: `<MagicMock name='mock.fget()' id='2183984899984'>`
The comparison with `self._scalar.value` used to simply give `False`, since they are clearly not identical. In Numpy 2, the result is ambiguous and gives an error
I can't quite tell if we need to update `value `to check if `_callback.fget()` actually produces a number, since it seems to only be an issue when mocking.
I've updated the tests to mock `_callback.fget()` as needed, but perhaps it's better to do it more generally?

### DIFF
--- a/.github/workflows/documentation-build.yml
+++ b/.github/workflows/documentation-build.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.9
+        python-version: 3.10
     - name: Install Pandoc, repo and dependencies
       run: |
         sudo apt install pandoc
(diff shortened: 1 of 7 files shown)


rozyczko (approved):
The changes look benign, and do what needs to be done.
Agreed with the mocking issue - this is pre-existing and we need to find a general solution for proper mocking of those cases.

PR #94, opened 2025-01-17
Remove old parameter

### LINKED ISSUE #14: Move to numpy 2
Numpy 2.0.0rc2 is almost ready for the release.
This version carries a number of API changes which are code breaking.
https://numpy.org/devdocs/numpy_2_0_migration_guide.html

First issue which shows up when trying to run pytest on easyscience is `np.cumproduct` which is now `np.cumpr`.
Likely, there are other differences.

The same issue needs propagating to other easyscience dependent modules.

### LINKED ISSUE #23: Move to new implementation of Descriptors and Parameters
Required clean when moving to the new structure of Descriptors and Parameters.

First we need to confirm that the new descriptors a parameters are working in:

- [ ] EasyDiffraction 
- [x] EasyReflectometry (https://github.com/EasyScience/EasyReflectometryLib/tree/easyscience-new-fitting-and-parameter)

To clean the code base:

search for 
- `.Variable import Descriptor` and delete
- `.Variable import Parameter` and delete

delete legacy tests:
- tests/integration_tests/Fitting/test_fitter_legacy_parameter.py
- tests/integration_tests/Fitting/test_multi_fitter_legacy_parameter.py
- tests/unit_tests/Fitting/test_constraints_legacy_parameter.py

rename folder: new_variable -> variable

Following files should be cleaned:
- pyproject.toml, pint is no longer a requirement
- src/easyscience/Objects/Inferface.py, clean up TODO sections
- src/easyscience/Objects/ObjectClasses.py, clean up TODO sections
- src/easyscience/fitting/Constraints.py, clean up TODO sections
- src/easyscience/fitting/minimizers/minimizer_bumps.py, clean up TODO sections
- src/easyscience/fitting/minimizers/minimizer_lmfit.py, clean up TODO sections
- tests/unit_tests/Fitting/test_constraints_legacy_parameter.py, delete
-tests/unit_tests/Objects/new_variable/test_descriptor_from_legacy.py, rename
- tests/unit_tests/Objects/new_variable/test_parameter_from_legacy.py, rename

Probably able to move to numpy 2

henrikjacobsenfys:
Removing the old parameter completely, renaming the `new_variable ` to `variable`. Added undo/redo functionality for converting units.


src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -44,8 +45,7 @@ def __init__(
         param url: URL of the descriptor
         param display_name: Display name of the descriptor
         param parent: Parent of the descriptor
-
-        .. note:: Undo/Redo functionality is implemented for the attributes `variance` and `value`.
+        .. note:: Undo/Redo functionality is implemented for the attributes `variance`, `error`, `unit` and `value`.
```
rozyczko, line 48: I don't think we have `error` anymore - it's `variance` now, right?

src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -192,9 +194,25 @@ def error(self, value: float) -> None:
...
+
+    #     :param unit_str: New unit in string form
+    #     """
+    #     if not isinstance(unit_str, str):
+    #         raise TypeError(f'{unit_str=} must be a string representing a valid scipp unit')
+    #     try:
+    #         new_unit = sc.Unit(unit_str)
+    #     except UnitError as message:
+    #         raise UnitError(message) from None
+    #     self._scalar = self._scalar.to(unit=new_unit)
```
rozyczko: please remove the commented out code

src/easyscience/Objects/variable/parameter.py
```diff
@@ -76,12 +76,14 @@ def __init__(
         :param parent: The object which is the parent to this one
 
         .. note::
-            Undo/Redo functionality is implemented for the attributes `value`, `error`, `min`, `max`, `fixed`
-        """
+            Undo/Redo functionality is implemented for the attributes `value`, `variance`, `error`, `min`, `max`, `bounds`, `fixed`, `unit`
```
rozyczko, line 79: likewise - `error` became `variance` so we don't need to include it

src/easyscience/Objects/variable/parameter.py
```diff
@@ -92,6 +94,7 @@ def __init__(
         if not isinstance(fixed, bool):
             raise TypeError('`fixed` must be either True or False')
 
+        self._fixed = fixed # For fitting, but must be initialized before super().__init__
```
rozyczko, line 97: there is nothing about `._fixed` in `DescriptorNumber`, which is the base class. Why does `self._fixed` need to be initialized here?

src/easyscience/Objects/virtual.py
```diff
@@ -16,6 +16,12 @@
 
 from easyscience import global_object
 from easyscience.Constraints import ObjConstraint
+from easyscience.Objects.variable.descriptor_base import DescriptorBase
```
rozyczko, line 19: not reviewing this file, since it is going to be deleted, so as long as the tests pass, I assume it is fine.

tests/unit_tests/global_object/test_undo_redo.py
```diff
@@ -59,6 +65,53 @@ def getter(_obj, _attr):
...
+)
+
+def test_DescriptorNumberUndoRedo(test):
+    obj = DescriptorNumber('DescriptorNumber',1,unit='m/s')
+    attr = test[0]
+    value = test[1]
+
+    e = doUndoRedo(obj, attr, value)
+    if e:
+        raise e
```
rozyczko: This is a weird test. I can see the original `test_SinglesUndoRedo` is written this way, but can't think about why do we raise here instead of just
`assert not e`
which will assure `doUndoRedo` worked as expected.
Do you know why we raise here explicitly?

rozyczko (approved):
The code looks good, several very minor issues raised.
Once resolved, this can be merged.

src/easyscience/Objects/variable/descriptor_number.py
henrikjacobsenfys: We still use `error`. We don't _need_ it, since it's just the square root of `variance `, but it's used in `repr`.

src/easyscience/Objects/variable/parameter.py
henrikjacobsenfys: It's complicated. `descriptor_number `calls `convert_unit` in `__init__`(*). This pushes a change to the undo/redo stack, which in its `__init__` makes an f string, which calls `__repr__`. `__repr__` of parameter checks if the parameter is fixed, and returns `self._fixed`. If it doesn't exist, we get an error.

(*) This is a bit of a hack, to make sure the units are what they're supposed to be. In earlier discussions, we had issues with units having numbers in them, such as a unit being `10 cm` or similar. The conversion fixes this issue.

tests/unit_tests/global_object/test_undo_redo.py
henrikjacobsenfys: No idea why. I'll change it to `assert not e` instead.

PR #99, opened 2025-01-29
Descriptor any type

henrikjacobsenfys:
Added a descriptor that can take any type.

I accidentally based this branch off of the DescriptorArray branch, but only DescriptorAnyType and its tests have been modified, so it should be fine.


rozyczko (approved):
Looks good. You might want to expand on the `DescriptorAnyType` docstring to mention the reason behind creating this class as it may be useful for the future refactoring.

PR #101, opened 2025-02-21
Updating bindings throws errors on value query on Descriptors

rozyczko:
Noticed when working on EDA.

This is a part of the large(r) slew of PRs for the new EasyDiffraction App update


henrikjacobsenfys:
So since this wasn't caught by any tests, I suppose we should write the test that should have caught it?

rozyczko:
> So since this wasn't caught by any tests, I suppose we should write the test that should have caught it?

There are no tests for InterfaceFactoryTemplate at all. I started looking at it but this is a separate project - we need proper coverage for this class, not just one test.

Delegated to https://github.com/easyscience/corelib/issues/102

PR #105, opened 2025-03-03
DescriptorArray

### LINKED ISSUE #103: ADR Suggestion: DescriptorArray
### Discussed in https://github.com/easyscience/corelib/discussions/95

`DescriptorArray` should be container for holding a set of independent `Descriptors`, built on `scipp.Variable`s. The class shall extend from `DescriptorBase`, and implements the following dunder methods (including their reverse operations)

- [x] Add
- [x] Sub
- [x] Mul
- [x] truediv
- [x] pow
- [x] abs
- [x] neg
- [x] getitem

Additionally, the following convenience functions shall be implemented
- [x] Sum
- [x] Trace

Note that we also want to add matrix multiplication functionality to `DescriptorArray`, but this is currently blocked by `scipp` not supporting general linear algebra operations. Furthermore, the `DescriptorArray` is not compatible with `numpy` methods.

elindgren:
Implements the `DescriptorArray` type, defined in #95 and #103 

Fixes #103


src/easyscience/Objects/variable/__init__.py
```diff
@@ -1,11 +1,11 @@
-from .descriptor_any_type import DescriptorAnyType
```
damskii9992, line 1: You can't remove the "DescriptorAnyType" from here as it is still being used in EasyCrystallography

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        :return: DescriptorArray
+        """
+        if not isinstance(full_value, Variable):
+            raise TypeError(f'{full_value=} must be a scipp array')
+        return cls(name=name, value=full_value.values, unit=full_value.unit, variance=full_value.variances, **kwargs)
+
+    @property
+    def full_value(self) -> Variable:
+        """
+        Get the value of self as a scipp array. This is should be usable for most cases.
```
damskii9992: typo "this should be"

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        The dict contains the value, unit and variances, in addition
+        to the properties of DescriptorBase.
+        """
+        raw_dict = super().as_dict(skip=skip)
+        raw_dict['value'] = self._array.values
+        raw_dict['unit'] = str(self._array.unit)
+        raw_dict['variance'] = self._array.variances
+        return raw_dict
+    
+    def _smooth_operator(self,
```
damskii9992: while surely a fun name, please use a descriptive method name for production-ready code :)

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+            # elements of the array. Hence, the broadcasting is "manually"
+            # performed to work around `scipp` and a warning raised to the end user.
+            if (self._array.variances is not None or other.variance is not None):
+                warn('Correlations introduced by this operation will not be considered.\
+                      See https://content.iospress.com/articles/journal-of-neutron-research/jnr220049\
+                      for further detailes', UserWarning)
+            # Cheeky copy() of broadcasted scipp array to force scipp to perform the broadcast here
+            broadcasted = sc.broadcast(other_converted.full_value, 
+                                             dims=self._array.dims,
+                                             shape=self._array.shape).copy()
```
damskii9992: why the copy? the `full_value` of a DescriptorNumber should be a Scipp Variable, and in their API reference, the broadcast should be a regular copy in this case. Only in the case of DataArray's is it a shallow copy.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+            new_full_value = operator(self.full_value, other_converted.full_value)
+
+        else:
+            return NotImplemented
+
+        descriptor_array = DescriptorArray.from_scipp(name=self.name, full_value=new_full_value)
+        descriptor_array.name = descriptor_array.unique_name
+        return descriptor_array
+
+    def _rsmooth_operator(self,
```
damskii9992: And same with the method name here

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+                          operator: str,
+                          units_must_match: bool = True) -> DescriptorArray:
+        """
+        Handle reverse operations for DescriptorArrays, DescriptorNumbers, lists, and scalars.
+        Ensures unit compatibility when `other` is a DescriptorNumber.
+        """
+        def reversed_operator(a, b):
+            return operator(b, a)
+        if isinstance(other, DescriptorArray):
+            # This is probably never called
```
damskii9992: Indeed, this will never be called. You can remove these 3 lines :)

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        """
+        raw_dict = super().as_dict(skip=skip)
+        raw_dict['value'] = self._array.values
+        raw_dict['unit'] = str(self._array.unit)
+        raw_dict['variance'] = self._array.variances
+        return raw_dict
+    
+    def _smooth_operator(self,
+                         other: Union[DescriptorArray, DescriptorNumber, list, numbers.Number],
+                         operator: str,
```
damskii9992: This type hint for operator is wrong, no?

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        Modify this function if you want to add such functionality.
+        """
+        return NotImplemented
+    
+    def __add__(self, other: Union[DescriptorArray, DescriptorNumber, list, numbers.Number]) -> DescriptorArray:
+        """
+        Perform element-wise addition with another DescriptorNumber, DescriptorArray, list, or number.
+
+        :param other: The object to add. Must be a DescriptorArray or DescriptorNumber with compatible units,
+                    or a list with the same shape if the DescriptorArray is dimensionless.
```
damskii9992: same shape or a number*

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        :return: A new DescriptorArray representing the result of the addition.
+        """
+        return self._smooth_operator(other, op.add)
+
+    def __radd__(self, other: Union[DescriptorArray, DescriptorNumber, list, numbers.Number]) -> DescriptorArray:
+        """
+        Handle reverse addition for DescriptorArrays, DescriptorNumbers, lists, and scalars.
```
damskii9992, line 505: You can safely remove DescriptorArray from here as it would never be called

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        :param other: The object to use as a denominator. Must be a DescriptorArray or DescriptorNumber with compatible units,
+                    or a list with the same shape if the DescriptorArray is dimensionless.
+        :return: A new DescriptorArray representing the result of the addition.
+        """
+        if not isinstance(other, (DescriptorArray, DescriptorNumber, list, numbers.Number)):
+            return NotImplemented
+
+        if isinstance(other, numbers.Number):
+            original_other = other
+        elif isinstance(other, (numbers.Number, list)):
```
damskii9992: why numbers.Number here? it is redundant and covered by the first if statement

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+            original_other = np.array(other)
+        elif isinstance(other, DescriptorNumber):
+            original_other = other.value
+        elif isinstance(other, DescriptorArray):
+            original_other = other.full_value.values
```
damskii9992: use the `value` method rather than `full_value.values`

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        return self._array
+
+    @full_value.setter
+    def full_value(self, full_value: Variable) -> None:
+        raise AttributeError(
+            f'Full_value is read-only. Change the value and variance seperately. Or create a new {self.__class__.__name__}.'
+        )
+
+    @property
+    def value(self) -> numbers.Number:
```
damskii9992, line 144: Should probably be renamed `values`

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+    def full_value(self) -> Variable:
+        """
+        Get the value of self as a scipp array. This is should be usable for most cases.
+
+        :return: Value of self with unit.
+        """
+        return self._array
+
+    @full_value.setter
+    def full_value(self, full_value: Variable) -> None:
```
damskii9992, line 138: Should probably be renamed to `full_values` i.e. add an "s"

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        if isinstance(value, list):
+            value = np.array(value)  # Convert lists to numpy arrays for consistent handling.
+
+        if value.shape != self._array.values.shape:
+            raise ValueError(f"{value=} must have the same shape as the existing array values.")
+
+        self._array.values = value
+    
+    @property
+    def dims(self) -> list:
```
damskii9992: I know we rely on Scipp and they call it `dims`, but I prefer to write stuff out and be explicit rather than save a few keystrokes. Would rename this to `dimensions`

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+    def unit(self, unit_str: str) -> None:
+        raise AttributeError(
+            (
+                f'Unit is read-only. Use convert_unit to change the unit between allowed types '
+                f'or create a new {self.__class__.__name__} with the desired unit.'
+            )
+        )  # noqa: E501
+
+    @property
+    def variance(self) -> float:
```
damskii9992: More plurals

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        descriptor_number.name = descriptor_number.unique_name
+        return descriptor_number
+
+    def __rpow__(self, other: numbers.Number) -> numbers.Number:
+        """
+        Defers reverse pow with a descriptor array, `a ** array`.
+        Exponentiation with regards to an array does not make sense,
+        and is not implemented.
+        """
+        return NotImplemented
```
damskii9992: I would probably raise an error here rather than return `NotImplemented`

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        Replace all elements in the DescriptorArray with their
+        absolute values. Note that this is different from the
+        norm of the DescriptorArray.
+        """
+        new_value = abs(self.full_value)
+        descriptor_array = DescriptorArray.from_scipp(name=self.name, full_value=new_value)
+        descriptor_array.name = descriptor_array.unique_name
+        return descriptor_array
+
+    def __getitem__(self, a) -> Union[DescriptorArray]:
```
damskii9992: why Union here? It's a single return type :)

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        trace = sc.scalar(0.0, unit=self.unit, variance=None)
+        for i in range(N):
+            # Index through all the dims to get
+            # the value i on the diagonal
+            diagonal_element = self.full_value
+            for dim in self.full_value.dims:
+                diagonal_element = diagonal_element[dim, i]
+            trace = trace + diagonal_element
+
+        descriptor = DescriptorNumber.from_scipp(name=self.name, full_value=trace)
```
damskii9992: Rename `descriptor` to `descriptor_number`, just to be consistent

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        descriptor.name = descriptor.unique_name
+        return descriptor
+
+    def sum(self, dim: Optional[Union[str, list]] = None) -> DescriptorNumber:
```
damskii9992: The return type type hinting is not correct, as the return type can also be a DescriptorArray

tests/unit_tests/Objects/variable/test_descriptor_array.py
```diff
@@ -0,0 +1,947 @@
...
+        assert descriptor._array.unit == "m"
+        assert descriptor._array.variances == None
+
+    # @pytest.mark.parametrize("full_value", [sc.array(values=[1,2], dims=["x"]), sc.array(values=[[1], [2]], dims=["x","y"]), object(), 1, "string"], ids=["1D", "2D", "object", "int", "string"])
+    # def test_from_scipp_type_exception(self, full_value):
+    #     # When Then Expect
+    #     with pytest.raises(TypeError):
+    #         DescriptorArray.from_scipp(name="name", full_value=full_value)
+
+    def tvigateDownest_full_value(self, descriptor: DescriptorArray):
```
damskii9992: what the hell is this test called? XD

tests/unit_tests/Objects/variable/test_descriptor_array.py
```diff
@@ -0,0 +1,947 @@
...
+        with pytest.raises(exception):
+            result = descriptor ** 2 ** test
+        with pytest.raises(TypeError):
+            # Exponentiation with an array does not make sense
+            test ** descriptor  
+
+    @pytest.mark.parametrize("test", [
+        DescriptorNumber("test", 2, "s"),
+        DescriptorArray("test", [[1, 2], [3, 4]], "s")], ids=["add_array_to_unit", "incompatible_units"])
+    def test_operation_exception(self, descriptor: DescriptorArray, test):
```
damskii9992: I would split this test into two to increase readability

tests/unit_tests/Objects/variable/test_descriptor_array.py
```diff
@@ -0,0 +1,947 @@
...
+                          [0.3 * 2**2 / 3**4, 0.4 * 2.0**2 / 4.0**4]]),
+         False)
+        ],
+        ids=["descriptor_number_regular",
+             "descriptor_number_unit_conversion",
+             "descriptor_number_different_units",
+             "array_conversion",
+             "list",
+             "number"])
+    def test_division(self, descriptor: DescriptorArray, test, expected, raises_warning):
```
damskii9992: I would not combine reverse tests with regular tests. Remember, you're testing a method, each unit test should ideally only test a single method for fine-grained error messages. Here you're testing two methods in a single test, even though they rely on the same `smooth_operator` method behind the scenes.

damskii9992 (changes requested):
I don't really know if I'm in favour of the `smooth_operator` implementation. More verbose code is sometimes better if it produces more readable code, but we should also get another persons opinion on this.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
+from __future__ import annotations
+
+import numbers
+import operator as op
```
rozyczko: hiding "internal" name is prone to errors, especially when you start using `operator` as a string later on

src/easyscience/Objects/variable/descriptor_array.py
rozyczko: This, again, is confusing - you are using Python's `operator` function, but renamed it to `op`, and here you have a string (or a function pointer?)

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        string += ">"
+        string=string.replace('\n', ',')
+        return string    
+
+
+    def as_dict(self, skip: Optional[List[str]] = None) -> Dict[str, Any]:
+        """
+        Dict representation of the current DescriptorArray.
+        The dict contains the value, unit and variances, in addition
+        to the properties of DescriptorBase.
```
rozyczko: our `ruff` specs allow lines up to 127 characters. No need to split unnecessarily.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        if not len(dims) == len(value.shape):
+            raise ValueError(f"Length of dims ({dims=}) does not match length of value {value=}.")
+        self._dims = dims
+
+        try:
+            self._array = sc.array(dims=dims, values=value, unit=unit, variances=variance)
+        except Exception as message:
+            raise UnitError(message)
+                # TODO: handle 1xn and nx1 arrays
+        self._array = sc.array(dims=dims, values=value, unit=unit, variances=variance)
```
rozyczko: this assignment is redundant (already done above)

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        Convert the value from one unit system to another.
+
+        :param unit_str: New unit in string form
+        """
+        if not isinstance(unit_str, str):
+            raise TypeError(f'{unit_str=} must be a string representing a valid scipp unit')
+        try:
+            new_unit = sc.Unit(unit_str)
+        except UnitError as message:
+            raise UnitError(message) from None
```
rozyczko: no additional context for this reraise - please consider either adding more info or let it percolate up

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        :param other: The object to use as a denominator. Must be a number or DescriptorNumber with
+                    no unit or variance.
+        :return: A new DescriptorArray representing the result of the addition.
+        """
+        if not isinstance(other, (numbers.Number, DescriptorNumber)):
+            return NotImplemented
+
+        if isinstance(other, numbers.Number):
+            exponent = other
+        elif type(other) is DescriptorNumber:
```
rozyczko: why not 
```python
elif isinstance(other, DescriptorNumber):
```

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        Get the value of self as a scipp array. This is should be usable for most cases.
+
+        :return: Value of self with unit.
+        """
+        return self._array
+
+    @full_value.setter
+    def full_value(self, full_value: Variable) -> None:
+        raise AttributeError(
+            f'Full_value is read-only. Change the value and variance seperately. Or create a new {self.__class__.__name__}.'
```
rozyczko: `seperately` -> `separately`

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+            # Operations with a DescriptorNumber that has a variance WILL introduce
+            # correlations between the elements of the DescriptorArray.
+            # See, https://content.iospress.com/articles/journal-of-neutron-research/jnr220049
+            # However, DescriptorArray does not consider the covariance between
+            # elements of the array. Hence, the broadcasting is "manually"
+            # performed to work around `scipp` and a warning raised to the end user.
+            if (self._array.variances is not None or other.variance is not None):
+                warn('Correlations introduced by this operation will not be considered.\
+                      See https://content.iospress.com/articles/journal-of-neutron-research/jnr220049\
+                      for further detailes', UserWarning)
```
rozyczko: `detailes` -> `details`

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,760 @@
...
+        """
+        return self.full_value.__delitem__(a)
+    
+    def __setitem__(self, a, b: Union[numbers.Number, list, DescriptorNumber, DescriptorArray]):
+        """
+        __setitem via slice is not allowed, since we currently do not give back a
+        view to the DescriptorArray upon calling __getitem__. 
+        """
+        raise AttributeError(
+            f'{self.__class__.__name__} cannot be edited via slicing. Edit the underlyinf scipp\
```
rozyczko: `underlyinf` -> `underlying`

rozyczko (commented):
Very well written code, good comments and adherence to our style.
Minor issues raised for consideration.

I like `smooth_operator` - it is internal, so no API will see it.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: The issue I had was that the returned object was not broadcast at this line without the `copy`. Rather, it was done as part of line 407 during the call `operator(self.full_value, broadcasted)`, thus running into the error that broadcasting during operations is not allowed in scipp:

```bash
scipp._scipp.core.VariancesError: Cannot broadcast object with variances as this would introduce unhandled correlations. 
```

src/easyscience/Objects/variable/descriptor_array.py
elindgren: I couldn't resist ;) Changed to `apply_operator`.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: I couldn't resist ;) Changed to `apply_operator`.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: Yes, it should be `Callable`. Fixed

src/easyscience/Objects/variable/descriptor_array.py
elindgren: I debated this myself, but I came to the conclusion that the `value` of a `DescriptorArray` would be it's underlying list. That is, I see an array as a single object, or value. `values` would imply a set of objects, which would be a collection of numbers (which a list of course can be seen as). Handling the array as a single value also makes sense in terms of the matrix operations that we've added/want to add, like `sum` and `trace`. Finally, I think it makes sense to have the same property names as for the other `Descriptors` as well.

tests/unit_tests/Objects/variable/test_descriptor_array.py
elindgren: no idea, it's one of Henrik's original tests xd

src/easyscience/Objects/variable/descriptor_array.py
elindgren: I forgot to update the type hint, it's supposed to be a `Callable`. Fixed

src/easyscience/Objects/variable/descriptor_array.py
elindgren: See above

src/easyscience/Objects/variable/descriptor_array.py
elindgren: See above

tests/unit_tests/Objects/variable/test_descriptor_array.py
elindgren: I see your point and generally agree. It's just gonna be a lot of very similar tests though, especially since the fixtures for the expected results will be copied. There are maybe things that one could do with wrapping the fixtures in fixtures themselves, but the test suite is already very long.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+    """
+    A `Descriptor` for Array values with units. The internal representation is a scipp array.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        value: Union[list, np.ndarray],
+        unit: Optional[Union[str, sc.Unit]] = '',
+        variance: Optional[numbers.Number] = None,
```
henrikjacobsenfys:
```suggestion
        variance: Optional[Union[list, np.ndarray]] = None,
```

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+            dimensions = ['dim'+str(i) for i in range(len(value.shape))]
+        if not len(dimensions) == len(value.shape):
+            raise ValueError(f"Length of dimensions ({dimensions=}) does not match length of value {value=}.")
+        self._dimensions = dimensions
+
+        try:
+            self._array = sc.array(dims=dimensions, values=value, unit=unit, variances=variance)
+        except Exception as message:
+            raise UnitError(message)
+                # TODO: handle 1xn and nx1 arrays
```
henrikjacobsenfys, line 94: Do we have a discussion somewhere of exactly how we want to handle them? It seems to work fine as it is.

src/easyscience/Objects/variable/descriptor_array.py
henrikjacobsenfys: I agree with using `value `rather than `values`.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+        raise AttributeError(
+            f'Full_value is read-only. Change the value and variance separately. Or create a new {self.__class__.__name__}.'
+        )
+
+    @property
+    def value(self) -> numbers.Number:
+        """
+        Get the value. This should be usable for most cases. The full value can be obtained from `obj.full_value`.
+
+        :return: Value of self with unit.
```
henrikjacobsenfys:
```suggestion
        :return: Value of self without unit.
```

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+    @full_value.setter
+    def full_value(self, full_value: Variable) -> None:
+        raise AttributeError(
+            f'Full_value is read-only. Change the value and variance separately. Or create a new {self.__class__.__name__}.'
+        )
+
+    @property
+    def value(self) -> numbers.Number:
+        """
+        Get the value. This should be usable for most cases. The full value can be obtained from `obj.full_value`.
```
henrikjacobsenfys:
```suggestion
        Get the value without units. The full scipp array can be obtained from `obj.full_value`.
```

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+    def dimensions(self) -> list:
+        """
+        Get the dimensions used for the underlying scipp array.
+
+        :return: dimensions of self.
+        """
+        return self._dimensions
+
+    @dimensions.setter
+    def dimensions(self, dimensions: Union[list, np.ndarray]) -> None:
```
henrikjacobsenfys: This doesn't actually change the names of the dimensions in the scipp array. Also, shouldn't they be a list of strings? E.g.
```
b=DescriptorArray(name='test',value=[[1],[2],[3]],unit='m') 
b.dimensions
['dim0', 'dim1']
```

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+            (
+                f'Unit is read-only. Use convert_unit to change the unit between allowed types '
+                f'or create a new {self.__class__.__name__} with the desired unit.'
+            )
+        )  # noqa: E501
+
+    @property
+    def variance(self) -> float:
+        """
+        Get the variance.
```
henrikjacobsenfys:
```suggestion
        Get the variance as an np array without units.
```

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+        obj_unit = str(self._array.unit)
+        if obj_unit and obj_unit != "dimensionless":
+            string += f", unit={obj_unit}"
+
+        string += ">"
+        string=string.replace('\n', ',')
+        return string    
+
+
+    def as_dict(self, skip: Optional[List[str]] = None) -> Dict[str, Any]:
```
henrikjacobsenfys, line 360: I suppose this should also save dimensions, or we can't reconstruct the array as it was from the dict, right?

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+            if isinstance(other, list):
+                # Use numpy to negate all elements of the list
+                value = (-np.array(other)).tolist()
+            else:
+                value = -other
+            return -(self.__radd__(value))
+        else:
+            return NotImplemented
+    
+    def __mul__(self, other: Union[DescriptorArray, DescriptorNumber, list, numbers.Number]) -> DescriptorArray:
```
henrikjacobsenfys, line 547: There's a problem sometimes with multiplication. I think it's because the numbers are integers in the following example. 
```
b=DescriptorArray(name='test',value=[[1],[2],[3]],unit='m')
c=DescriptorArray(name='test',value=[[1],[2],[3]],unit='cm')
b*c
<DescriptorArray 'DescriptorArray_23': values=[[0], [0], [0]], unit=m^2>
```

src/easyscience/Objects/variable/descriptor_array.py
elindgren: We investigated the issue some more, and it seems like `copy()` is needed since `broadcast` returns a view into the array. In other words, the underlying data is not duplicated, and hence we need to add the copy to force a deep copy into actually separate values.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+    def __rmul__(self, other: Union[DescriptorNumber, list, numbers.Number]) -> DescriptorArray:
+        """
+        Handle reverse multiplication for DescriptorNumbers, lists, and scalars.
+        Ensures unit compatibility when `other` is a DescriptorNumber.
+        """
+        if not isinstance(other, (DescriptorNumber, list, numbers.Number)):
+            return NotImplemented
+        return self._rapply_operation(other, operator.mul, units_must_match=False)
+    
+    def __truediv__(self, other: Union[DescriptorArray, DescriptorNumber, list, numbers.Number]) -> DescriptorArray:
```
henrikjacobsenfys, line 568: Similar problems here

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,754 @@
...
+        Computes the trace over the descriptor array.
+        Only works for matrices where all dimensions are equal.
+        """
+        shape = np.array(self.full_value.shape)
+        N = shape[0]
+        if not np.all(shape == N):
+            raise ValueError('\
+                    Trace can only be taken over arrays where all dimensions are of equal length')
+
+        trace = sc.scalar(0.0, unit=self.unit, variance=None)
```
henrikjacobsenfys:
```suggestion
        trace = sc.scalar(np.trace(self.value), unit=self.unit, variance=None)
```

henrikjacobsenfys (commented):
Well done! Just a few issues.

src/easyscience/Objects/variable/descriptor_array.py
damskii9992: Yes, this is actually a critical point I had missed. Integers need to be converted into floats in the constructor like we do for DescripturNumber.

src/easyscience/Objects/variable/descriptor_array.py
damskii9992: Doing it this way we don't get the automatic error propagation. I'm pretty sure thats why Eric is doing it this other way :)

tests/unit_tests/Objects/variable/test_descriptor_array.py
elindgren: Tests have been split.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: I removed the try catch block. I took the block from `DescriptorNumber`, but it seems like `scipp` has been updated so this is not needed anymore.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: No discussion as far as I know, this comment was here when I started working on it. Yes, in my tests I try 1xn and nx1 arrays, and they work fine.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: fixed

src/easyscience/Objects/variable/descriptor_array.py
henrikjacobsenfys: Haha that's my comment then, from when I hard coded it to be n  x m matrices.

src/easyscience/Objects/variable/descriptor_array.py
elindgren: fixed

src/easyscience/Objects/variable/descriptor_array.py
elindgren: yes, that is indeed the case

henrikjacobsenfys: Of course, my bad. I was fooled by the variance=None when it was initialised, thinking it wasn't relevant.

elindgren:
I think that should be everything taken care of! We can merge if you guys are happy imo

Edit: @damskii9992 re-requested the review from you since you requested changes.

src/easyscience/Objects/variable/descriptor_array.py
henrikjacobsenfys: We could of course compute the variance as the trace of the variance array, it may be cleaner?

elindgren: Good idea @henrikjacobsenfys, that works very nicely! I found another error in my `trace` implementation. I thought that the trace of a tensor would always be a number, but that doesn't seem to be the case. In general, a tensor trace of a rank `k` tensor is of rank `k-2`. So, in Einstein summation notation,

$$
\begin{align}
    \mathrm{tr}(A) &= A^i_{i}, \hspace{10px} \mathrm k=2 \\
    \mathrm{tr}(A) &= A^i_{im}, \hspace{10px} \mathrm k=3
\end{align}
$$

etc. This is also what numpy yields. I suppose this is what we want to implement?

elindgren: I updated the `trace` implementation to use numpy as you suggested @henrikjacobsenfys. The implementation is way more readable imo.

elindgren: The trace function now also can take over which dimensions the trace should be computed.

src/easyscience/Objects/variable/descriptor_array.py
```diff
@@ -0,0 +1,793 @@
...
+            raise ValueError('Either both or none of `dimension1` and `dimension2` must be set.')
+        
+        if dimension1 is not None and dimension2 is not None:
+            if dimension1 == dimension2:
+                raise ValueError(f'`{dimension1=}` and `{dimension2=}` must be different.')
+            
+            axes = []
+            for dim in (dimension1, dimension2):
+                if dim not in self.dimensions:
+                    raise ValueError(f'Dimension {dim=} does not exist in the ')
```
damskii9992: Missing ending of string

damskii9992 (approved):
ValueError string is incomplete, but otherwise I would finally say this is ready to be merged :)

src/easyscience/Objects/variable/descriptor_array.py
elindgren: Fixed

PR #112, opened 2025-04-10
Issue34 dependent parameters

### LINKED ISSUE #34: Replace constraints with dependent Parameters
### Discussed in https://github.com/EasyScience/EasyScience/discussions/10

<sup>Originally posted by **damskii9992** June  7, 2024</sup>

# Requirements
- When an independent parameter value is changed, it should update the values of all the parameters which depend directly on it. When doing this update, signals should also be send to the GUI frontend to reflect this change.
- Making a parameter dependent should be easy and user-friendly, both when generated and when changed by the setter, by allowing direct expressions such as: 'value=a+2*b' in the value field
- A dependent parameter should also be easy to change back into an independent parameter
- Dependent parameters should not be included in the fitting procedure
- [Optional] It should be possible to visualize a dependency tree to get an overview of all parameter dependencies

# Implementation
- '_is_independent' boolean variable, for checks
- 'list_of_dependents' list variable with UIDs of parameters that depend on the parameter
- '_update' method, to be called on dependent parameters when their independent dependency is changed
- Logic to ensure no broken, cyclic or secondary (for now) dependencies</div>

damskii9992:


damskii9992:
Have to add a deprecated label for the CI to proceed

damskii9992:
Hmm, the `_process_dependency_symbol_names` actually does not work, because the output of `globals()` from a module yields the modules global scope, rather than the users. I do not see a solution to this problem and will instead try out an implementation with a dependency_map dictionary input

src/easyscience/Objects/variable/parameter.py
```diff
@@ -68,32 +71,55 @@ def __init__(
         :param variance: The variance of the value
         :param min: The minimum value for fitting
         :param max: The maximum value for fitting
-        :param fixed: Can the parameter vary while fitting?
+        :param fixed: Can the parameter vary?
         :param description: A brief summary of what this object is
         :param url: Lookup url for documentation/information
         :param display_name: The name of the object as it should be displayed
-        :param enabled: Can the objects value be set
+        :param independent: Is the object dependent on another object?
```
rozyczko: We use the `enabled` state quite extensively in the libs/guis. Why did you remove this attribute?

src/easyscience/Objects/variable/parameter.py
```diff
@@ -124,6 +149,22 @@ def __init__(
...
+        Update the parameter. This is called by the DescriptorNumbers/Parameters who have this Parameter as a dependency.
+        """
+        if not self._independent:
+            temporary_parameter = self._global_object.dependency_interpreter(self._clean_dependency_string)
+            self._scalar.value = temporary_parameter.value
+            self._scalar.unit = temporary_parameter.unit
+            self._scalar.variance = temporary_parameter.variance
+            self._min.value = temporary_parameter.min
+            self._max.value = temporary_parameter.max
+            self._notify_observers()
```
rozyczko: The only occurrence of this method being called is in `DescriptorNumber`. It is called from `_notify_observers()` method.
Does this mean the `self._notify_observers()` called here will be also called the DescriptorNumber and cause an infinite loop?

src/easyscience/Objects/variable/parameter.py
```diff
@@ -174,37 +215,62 @@ def value(self, value: numbers.Number) -> None:
...
+            if self._callback.fset is not None:
+                self._callback.fset(self._scalar.value)
 
-        self._scalar.value = float(value)
-        if self._callback.fset is not None:
-            self._callback.fset(self._scalar.value)
+            # Notify observers of the change
+            self._notify_observers()
+        else:
+            raise AttributeError("This parameter is not independent, its value cannot be set directly. Please make it independent first.")  # noqa: E501
```
rozyczko: Do we want to raise here? (And the methods below).
Maybe just notify the user and drop the assignment attempt.

rozyczko (commented):
General points:
- complexity: AST parsing and `asteval.Interpreter` pose significant complexity, making the code potentially harder to understand and maintain.
- performance: the observer pattern and dependency evaluation could introduce performance overhead, especially in scenarios with many dependent parameters or frequent updates. We need to run benchmarks to see if the overhead is not too serious
- error handling: too much reliant on exceptions, which can clutter logs, cause unwanted exits or fail to provide clear error messages in complex setups

Additionally, complex methods like `_process_dependency_symbol_names` and `_process_dependency_unique_names` need clear examples for developers to understand their usage.

src/easyscience/Objects/variable/parameter.py
damskii9992: Nope. Remember that `Parameter ` inherits from `DescriptorNumber`. The only reason the `_notify_observers()` methods is written in the `DescriptorNumber` code, is because a `Parameter` should be able to depend on a `DescriptorNumber`. 
Currently there can be an infinite loop if a dependent parameter is dependent on another dependent parameter which relies on itself, since each of those parameters have the other parameter in their `_observers` list, but this will be handled by my unique update id suggestion

src/easyscience/Objects/variable/parameter.py
damskii9992: Well, can you tell me what its purpose is?
How does it differ at all from "fixed" or (not) "independent"?
In all the code I could see about it, it looked like it was just an alias for not independent.

src/easyscience/Objects/variable/parameter.py
damskii9992: Well, we raise when the user tries to set the full value. Since that is a read-only property.
For a dependent parameter, the value (and other entries) are also read-only properties, so I figured they should be handled in the same way. We could also raise a warning, but trying to set the value of a dependent parameter is an error, and that is why I flag it as such.

damskii9992:
> General points:
> 
> * complexity: AST parsing and `asteval.Interpreter` pose significant complexity, making the code potentially harder to understand and maintain.

As I mentioned in my own comment, the AST parsing didn't work anyways due to the global namescope not being available inside modules, so I am currently removing it in favour of a simpler (albeit more verbose) approach where the user has to pass a dictionary of the dependencies along with the string to create a dependent parameter (unless unique_names are used).

> * performance: the observer pattern and dependency evaluation could introduce performance overhead, especially in scenarios with many dependent parameters or frequent updates. We need to run benchmarks to see if the overhead is not too serious

Yes, I think this is an issue we would run into with any implementation of dependent parameters, including our current one, but we will definitely need to benchmark it. Especially with the approach I am currently implementing where each Parameter gets its own asteval interpreter, I am concerned about the potential performance issue as I am unsure of how heavy the asteval interpreter object is.

> * error handling: too much reliant on exceptions, which can clutter logs, cause unwanted exits or fail to provide clear error messages in complex setups

What do you mean? Most of the error handling I do is specifically aimed at providing clearer error messages when stuff fails.

> Additionally, complex methods like `_process_dependency_symbol_names` and `_process_dependency_unique_names` need clear examples for developers to understand their usage.

I was gonna go through the implementation of the `_process_dependency_symbol_names` method in the ADR suggestion due to its complexity, with an example, before that implementation idea was scratched. 
Is the `_process_dependency_unique_names` a complicated method? I find it pretty self-explanatory, but maybe I've gotten to used to my own code to see it?

src/easyscience/Constraints.py
```diff
@@ -60,11 +60,11 @@ def __init__(
             # Test if dependent is a parameter or a descriptor.
             # We can not import `Parameter`, so......
             if dependent_obj.__class__.__name__ == 'Parameter':
-                if not dependent_obj.enabled:
+                if not dependent_obj.independent:
                     raise AssertionError('A dependent object needs to be initially enabled.')
```
elindgren:
```suggestion
                    raise AssertionError('A dependent object needs to be initially independent.')
```
Or does semantically make sense? What did it mean for a dependent_obj to be `enabled`?

src/easyscience/Constraints.py
```diff
@@ -60,11 +60,11 @@ def __init__(
             # Test if dependent is a parameter or a descriptor.
             # We can not import `Parameter`, so......
             if dependent_obj.__class__.__name__ == 'Parameter':
```
elindgren:
```suggestion
            if isinstance(dependent_obj, Parameter)
```

src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -229,6 +271,14 @@ def set_scalar(obj, scalar):
         # Update the scalar
         self._scalar = new_scalar
 
+    @notify_observers
+    def convert_unit(self, unit_str: str) -> None:
+        """
```
elindgren, line 279: This is the user-facing version of `convert_unit`, right? So from the user's perspective nothing has changed API-wise w.r.t converting units?

src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -198,7 +240,7 @@ def error(self, value: float) -> None:
         else:
             self._scalar.variance = None
 
-    def convert_unit(self, unit_str: str) -> None:
+    def _convert_unit(self, unit_str: str) -> None:
         """
         Convert the value from one unit system to another.
```
elindgren, line 247: Add a sentence motivating why both `_convert_unit` and `convert_unit` exists (to avoid cyclic deps).

src/easyscience/Objects/variable/parameter.py
```diff
@@ -68,11 +72,10 @@ def __init__(
         :param variance: The variance of the value
         :param min: The minimum value for fitting
         :param max: The maximum value for fitting
-        :param fixed: Can the parameter vary while fitting?
+        :param fixed: Can the parameter vary?
```
elindgren:
```suggestion
        :param fixed: If the parameter is free to vary during fitting
```

src/easyscience/Objects/variable/parameter.py
```diff
@@ -124,6 +126,166 @@ def __init__(
...
+                self._notify_observers(update_id=update_id)
+        else:
+            warnings.warn('This parameter is not dependent. It cannot be updated.')
+
+    def make_dependent(self, dependency_expression: str, dependency_map: Optional[dict] = None) -> None:
+        """
+        Make this parameter dependent on another parameter. This will overwrite the current value, unit, variance, min and max.
+
+        :param dependency_expression: The dependency expression to evaluate. This should be a string which can be evaluated by the ASTEval interpreter.
+        :param dependency_map: A dictionary of dependency expression symbol name and dependency object pairs. This is inserted into the asteval interpreter to resolve dependencies.
```
elindgren, line 175: Maybe add an example shape of what the dependency map looks like? I read this like the map contains multiple dependencies. But it's just a map of the form `{'a': a, 'b': b}`?

src/easyscience/Objects/variable/parameter.py
```diff
@@ -124,6 +126,166 @@ def __init__(
...
+
+        # If we're overwriting the dependency
+        if not self._independent:
+            for old_dependency in self._dependency_map.values():
+                old_dependency._detach_observer(self)
+
+        self._dependency_string = dependency_expression
+        self._dependency_map = dependency_map if dependency_map is not None else {}
+        self._dependency_interpreter = Interpreter(minimal=True)
+        self._dependency_interpreter.config['if'] = True
```
elindgren: maybe add a comment on why this needs to be set to True.

src/easyscience/Objects/variable/parameter.py
```diff
@@ -124,6 +126,166 @@ def __init__(
...
+        :return: True = independent, False = dependent
+        """
+        return self._independent
+    
+    @independent.setter
+    def independent(self, value: bool) -> None:
+        raise AttributeError('This property is read-only. Use `make_independent` and  `make_dependent` to change the state of the parameter.')  # noqa: E501
+
+    @property
+    def depedency_expression(self) -> str:
```
elindgren:
```suggestion
    def dependency_expression(self) -> str:
```

src/easyscience/Objects/variable/parameter.py
```diff
@@ -399,24 +605,30 @@ def _constraint_runner(
...
+        inputted_unique_names = re.findall("(\'.+?\')", dependency_expression)
+        inputted_unique_names += re.findall('(\".+?\")', dependency_expression)
 
-        :param value: True - objects value can be set, False - the opposite
-        """
-        self._enabled = value
+        clean_dependency_string = dependency_expression
+        existing_unique_names = self._global_object.map.vertices()
+        # Add the unique names of the parameters to the ASTEVAL interpreter
+        for name in inputted_unique_names:
```
elindgren, line 527: inputted_unique_names only contains global ids, right? So if no global id's are passed in the expression this will be skipped?

elindgren (commented):
I had a look at the current implementation. I think it is very readable actually, and it seems like a lot of cases have been handled w.r.t cleanup etc. My main concern is performance because of ASTeval, but I don't see a better way of doing things. Would be interesting to see what the computational cost is for a real-life example.

src/easyscience/Objects/variable/descriptor_number.py
damskii9992: Yes, exactly. Nothing has changed from the user-perspective.

src/easyscience/Objects/variable/parameter.py
damskii9992: Yes and no. Say your dependency expression is "a+b". 
Where parameter a is `cell_length_a` and parameter b is `cell_length_b`, both belonging to the model.
An example of a working dependency map would then be:
```python
dependency_map = {
   'a' : model.cell_length_a,
   'b' : model.cell_length_b
}
```
I.e. the values are any working reference to the actual object, whereas the keys are how they are represented in the dependency expression

src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -90,6 +109,26 @@ def from_scipp(cls, name: str, full_value: Variable, **kwargs) -> DescriptorNumb
...
+    def _attach_observer(self, observer: DescriptorNumber) -> None:
+        """Attach an observer to the descriptor."""
+        self._observers.append(observer)
+
+    def _detach_observer(self, observer: DescriptorNumber) -> None:
+        """Detach an observer from the descriptor."""
+        self._observers.remove(observer)
+
+    def _notify_observers(self, update_id=None) -> None:
+        """Notify all observers of a change.
```
rozyczko, line 121: At the only point this is called, the arguments are null. Consider getting rid of `update_id` as an argument. This will also remove the conditional in line 126.

src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -422,4 +472,4 @@ def _base_unit(self) -> str:
                     return string[i:]
             elif letter not in ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '.', '+', '-']:
                 return string[i:]
-        return ''
```
rozyczko, line 425: unrelated but you added a blank line at the end so I noticed this ugly loop with conditionals
Consider replacing it with a more compact version.
Or not, if you prefer code being more readable but potentially slower.
```
import re

def _base_unit(self) -> str:
    """
    Extract the base unit from the unit string by removing numeric components and scientific notation.
    """
    match = re.search(r'[^\d.eE+-]', str(self._scalar.unit))
    start = match.start()
    return str(self._scalar.unit)[start:] if match else ''
```

src/easyscience/Objects/variable/descriptor_number.py
```diff
@@ -115,6 +154,7 @@ def value(self) -> numbers.Number:
         return self._scalar.value
 
     @value.setter
+    @notify_observers
     @property_stack_deco
```
rozyczko: This decorator should really be renamed to `property_stack` only. `_deco` looks weird.
Unrelated but annoying

src/easyscience/Objects/variable/parameter.py
damskii9992: Yup. If there are no unique_names in the dependency expression, `inputted_unique_names` will be empty and therefore the for-loop will not run.

src/easyscience/Objects/variable/descriptor_number.py
damskii9992: Nope, check the `_update` method on the `Parameter` class, this is the critical point where it is called with an update_id.
This particular argument is how we avoid cyclic dependencies.

src/easyscience/Objects/variable/descriptor_number.py
damskii9992: Do we want regex syntax when we can avoid it?
Your proposed code is indeed shorter, but I find it harder to read. I am not sure what exactly that regex search matches.

src/easyscience/Objects/variable/descriptor_number.py
damskii9992: Yes, I agree. We can sneak that change into this PR :D

src/easyscience/Objects/variable/descriptor_number.py
damskii9992: Also, this regex search fails if the unit happens to start with an "e", as it will skip the "e" thinking it is scientific notation (which it might not be).

rozyczko: Yeah, regex is more obscure, but at least it avoids those unnecessary conditionals. Good.

docs/src/fitting/constraints.rst
rozyczko, line 1: Consider replacing rather than deleting the documentation on constraints.
It will be required to have this available and the best time to write the docs is when the feature is implemented...

damskii9992: I don't think these docs actually work or are published anywhere. At least the contents does not appear to be similar to https://easyscience.github.io/corelib/modules/constraints/

damskii9992: In fact, should I just remove the entire docs folder? There is basically nothing in it.
I still don't know exactly what our documentation plans are.

src/easyscience/Objects/variable/descriptor_number.py
elindgren: Nice!

src/easyscience/Objects/variable/parameter.py
```diff
@@ -115,14 +110,189 @@ def __init__(
...
+        self._scalar.value = dependency_result.value
+        self._scalar.unit = dependency_result.unit
+        self._scalar.variance = dependency_result.variance
+        self._min.value = dependency_result.min if isinstance(dependency_result, Parameter) else dependency_result.value
+        self._max.value = dependency_result.max if isinstance(dependency_result, Parameter) else dependency_result.value
+        self._min.unit = dependency_result.unit
+        self._max.unit = dependency_result.unit
+        self._independent = False
+        self._fixed = False
+        self._notify_observers()
```
seventil, line 231: shouldn't it be replaced with a new decorator?

src/easyscience/global_object/global_object.py
```diff
@@ -36,6 +36,8 @@ def __init__(self):
         # Map. This is the conduit database between all global object species
         self.map: Map = self.__map
 
+        self.update_id_iterator = 0
```
rozyczko, line 40: please describe this variable, seeing as every other self variable is properly documented.

src/easyscience/Objects/variable/parameter.py
damskii9992: I didn't realize it could actually be used here, :O  as it can't be used in most of the setters due to the check for independence. I don't know if it provides any benefit to use it though. Actually I think the code is slightly easier to read with the explicit `_notify_observers()` call.

tests/unit_tests/Objects/test_BaseObj.py
```diff
@@ -148,7 +148,7 @@ def test_baseobj_fit_objects(setup_pars: dict):
     pass
 
 
-def test_baseobj_as_dict(setup_pars: dict):
+def test_baseobj_as_dict(clear, setup_pars: dict):
```
elindgren, line 151: what does the argument (fixture?) `clear` do? I don't see it being called anywhere in your changes.

tests/unit_tests/Objects/variable/test_parameter.py
```diff
@@ -91,10 +113,326 @@ def test_init_value_max_exception(self):
...
+        assert normal_parameter._independent == False
+        assert normal_parameter.dependency_expression == '2*a'
+        assert normal_parameter.dependency_map == {'a': independent_parameter}
+        self.compare_parameters(normal_parameter, 2*independent_parameter)
+
+        # Then
+        independent_parameter.value = 2
+
+        # Expect
+        self.compare_parameters(normal_parameter, 2*independent_parameter)
```
elindgren, line 137: Maybe add a quick check to make sure that `normal_parameter.value == 4` explicitly here. It makes the test easier to read, and also make sure that the value of the independent parameter is actually updated and propagates.

tests/unit_tests/Objects/variable/test_parameter.py
```diff
@@ -91,10 +113,326 @@ def test_init_value_max_exception(self):
...
+
+    def test_dependent_parameter_with_unique_name(self, clear, normal_parameter: Parameter):
+        # When Then
+        dependent_parameter = Parameter.from_dependency(
+            name = 'dependent', 
+            dependency_expression='2*"Parameter_0"',
+        )
+
+        # Expect
+        assert dependent_parameter.dependency_expression == '2*"Parameter_0"'
```
elindgren, line 170: Now that I see this again, can you remind me why we can't simply do `2*Parameter_0"? First check if `Parameter_0` is defined in the map, and then check if it is a global parameter. 

It is not as unambiguous as using quotes to delimit the global parameters, but I'm afraid it would be unintuitive for users. I guess not many users would use the global_id anyway, so might be a nothing burger.

tests/unit_tests/Objects/variable/test_parameter.py
```diff
@@ -91,10 +113,326 @@ def test_init_value_max_exception(self):
...
+        assert normal_parameter._dependency_map == {'__Special_name__': independent_parameter, 
+                                                    '__Special_name_2__': independent_parameter_2}
+        assert normal_parameter._clean_dependency_string == '__Special_name__ + __Special_name_2__'
+
+    def test_process_dependency_unique_names_exception_unique_name_does_not_exist(self, clear, normal_parameter: Parameter):
+        # When
+        normal_parameter._dependency_map = {}
+
+        # Then Expect
+        with pytest.raises(ValueError):
```
elindgren:
```suggestion
        with pytest.raises(ValueError) as e: 
```
and then test the error message with `assert 'TEST string' in str(e)` perhaps

elindgren: Ah I see one can do it cleaner with `pytest.rases(..., match=)` as in your next test.

tests/unit_tests/Objects/variable/test_parameter.py
```diff
@@ -91,10 +113,326 @@ def test_init_value_max_exception(self):
...
+
+        # Then Expect
+        with pytest.raises(RuntimeError):
+            normal_parameter.make_dependent_on(dependency_expression='2*c', dependency_map={'c': dependent_parameter_2})
+
+    def test_dependent_parameter_logical_dependency(self, normal_parameter: Parameter):
+        # When
+        dependent_parameter = Parameter.from_dependency(
+            name = 'dependent',
+            dependency_expression='a if a.value > 0 else -a',
```
elindgren, line 325: very nice!

tests/unit_tests/Objects/variable/test_parameter.py
```diff
@@ -91,10 +113,326 @@ def test_init_value_max_exception(self):
...
+        # When
+        dependent_parameter = Parameter.from_dependency(
+            name = 'dependent',
+            dependency_expression='2*a',
+            dependency_map={'a': normal_parameter},
+        )
+
+        # Then Expect
+        with pytest.raises(AttributeError):
+            dependent_parameter.dependency_expression = '3*a'
```
elindgren, line 418: so not allowed to overwrite the dependency expression after the fact?

elindgren (commented):
I think the unit tests for the dependent parameters looks very thorough! Just had a few comments.

tests/unit_tests/Objects/test_BaseObj.py
damskii9992: The clear fixture clears out the `global_object`, to ensure that newly created objects unique_names doesn't crash with existing ones, since the interpreter is not reset between unit tests.

tests/unit_tests/Objects/variable/test_parameter.py
damskii9992: Well, that check is within the `compare_parameters` function. Do you think the test becomes less readable because I hide the checks away behind this function?

tests/unit_tests/Objects/variable/test_parameter.py
damskii9992: It could be an option. We'd have to parse the string then, possibly with ast, to get all the names in the expression, and then check if they're in the dependency_map keys and if not in the map, then check the global_object.

I don't think its worth it though, as it makes the code more complex to allow for this, and the usage is kinda unintuitive as well as you say. At least when the user encapsulates names with "" or '', then they KNOW it should be an unique_name (if they read the docs).

tests/unit_tests/Objects/variable/test_parameter.py
damskii9992: :D Do you think I should include the string match in this test?

tests/unit_tests/Objects/variable/test_parameter.py
damskii9992: Not directly no, the error message thrown tells the user to use the `make_dependent_on` method again if they wish to overwrite it.
This is so that the user can't easily break the dependency by for example changing the dependency expression to include names not in the map. Mostly it is because the dependency expression and map should be updated simultaneously :)

tests/unit_tests/Objects/test_BaseObj.py
elindgren: Okay! Is it being run by every test in the file? If so, consider converting it to an [`autouse` test](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request)

tests/unit_tests/Objects/variable/test_parameter.py
elindgren: No, I think using `compare_parameters` is good. What I mean was to, additionally, test that `assert normal_parameter.value == 4` on the line after the current check. It is a bit of a unnecessary test, as the only way it would fail is if the value of `independent_parameter` fails to update, which should be tested in that unit test, but I usually like to throw in explicit numeric checks here and there in my tests to make sure that the test is independent.

tests/unit_tests/Objects/variable/test_parameter.py
elindgren: Fair enough. Let's keep it as is.

tests/unit_tests/Objects/variable/test_parameter.py
elindgren: Yes, I think it is nice to ensure that the error one expects is actually being thrown.

tests/unit_tests/Objects/variable/test_parameter.py
elindgren: Ahhh I see, very good catch!

tests/unit_tests/Objects/test_BaseObj.py
damskii9992: No, it is not. It could be run for every test, as it wouldn't break anything, but currently it is only added to the tests where it is specifically needed. I suppose doing it this way is actually helpful, as it shows us which tests have issues with the `global_object`.

elindgren: Okay, fair point.

docs/src/fitting/constraints.rst
AndrewSazonov: I suggest keeping all the existing documentation for now, at least until I finish working on the new documentation templating system. Once the new template is in place, we can review the old content and see what can be reused or adapted. That would also be a good opportunity to clean things up.

src/easyscience/Objects/variable/parameter.py
AndrewSazonov: Trying to recall it... I believe we used the `enabled` attribute to indicate that a parameter is not constrained, meaning it is independent. If you’re introducing a new attribute called `independent`, that should be fine, as independent is more self-explanatory than enabled.

The `fixed` attribute is a separate matter. It was used for independent parameters to control whether they should be optimized during the fitting process. As we discussed previously, there was a suggestion to use `free` instead. The idea is that `par.free = True` is generally more intuitive and readable than `par.fixed = False`.

PR #120, opened 2025-05-07
Major clean up of source code and restructuring.

### LINKED ISSUE #116: Spring cleaning of unused source code
A large portion of our source-code is unused by any of our projects and should therefore be removed to ease maintenance.
Specifically:

- `unsafe_hash` method in the `ComponentSerializer` class
- `as_data_dict` method in the `ComponentSerializer` class
- `encode_data` method in the `ComponentSerializer` class
- `jsanatize` in `JsonDecoderTemplate` class
- class `EasyScienceDatasetAccessor` in xarray.py*
- class `EasyScienceDataarrayAccessor` in xarray.py*
- `check_sanity_single` in xarray.py*
- `check_sanity_multiple` in xarray.py*
- Entire "Datasets" folder including xarray.py
- Exceptions.py file
- class `XMLSerializer` in xml.py
- class `JsonDataSerializer` in json.py
- class `JsonSerializer` in json.py**
- class `DataDictSerializer` in dict.py
- function `designate_calc_fn` in polynomial.py
- class `Line` in polynomial.py
- `_get_class_module` in `SerializerBase` class

*Current version of EasyDiffraction (v0.4.0) uses it, but it is to be removed next release and converted to Scipp.
**It was decided to use JSON as our primary format for saving/storing sessions, yet this Serializer was not being used anywhere. Instead the regular `DataSerializer` was used together with `json.dump` from the json package.

We also have dependencies which are no longer used and can be removed:

- xarray
- pint

### LINKED ISSUE #118: Clean up source code folder structure
### Discussed in https://github.com/easyscience/corelib/discussions/114

<sup>Originally posted by **damskii9992** May  1, 2025</sup>
# General
The source code folder structure is important both from an user-experience point-of-view, but also for a maintainability point-of-view. When source code is structured into clearly distinct modules for their distinct functionalities, and when classes aren't hidden away in files with ambiguous names, it is easier for maintainers to find files and quickly gather what the purpose of the file is just from its location.

Good use of sub-modules also makes it easier for users to import their needed classes and functions and allows us to re-name things behind the scenes and provide aliases without directly interfering with the user-experience.

The current source code folder structure is confusing for both maintainers and users alike.

### LINKED ISSUE #123: `ADR Suggestion` Detecting cyclic dependencies easily with pings
### Discussed in https://github.com/easyscience/corelib/discussions/122

<sup>Originally posted by **damskii9992** May 16, 2025</sup>
# General
Since parameters can be made dependent after they're created, it is possible to end up in an infinite loop where a dependent parameters update triggers another dependent parameters update which in turn triggers the first parameters update and so on:
```mermaid
flowchart LR;
   A-->|update|B;
   B-->|update|C;
   C-->|update|D;
   D-->|update|B;
```
In ADR #10 a system for detecting cyclic dependencies were suggested and implemented. In this ADR I propose a simpler yet smarter system, which also properly handles the edge-case described in #10.

# Current Implementation
In the current implementation, the cyclic dependencies are discovered through the observer pattern with **update_ids**:
If an update is triggered by a manual change, such as a value change of an independent parameter, the update gets assigned a unique id.
This unique id is then passed to the observers along with the unique_name of the object that triggered the update;
```python
def _notify_observers(self, update_id=None) -> None:
     """Trigger an update in each subscriber."""
     if update_id is None:
        self._global_object.update_id_iterator += 1
        update_id = self._global_object.update_id_iterator
     for observer in self._observers:
         observer._update(update_id=update_id, updating_object=self.unique_name)
```
In the observers _update method, the update id is checked against an internal dictionary for the updating object. If the stored update id for the updating object is different from the new one, set the stored update id to the new one and proceed with the update, and notify its own observers, passing along the update_id, otherwise raise an error warning that a cyclic dependency has been detected:
```python
def __update(self, update_id: int, updating_object: str) -> None
   if self._dependency_updates[updating_object] == update_id:
      raise RuntimeError
   else:
      update stuff
      self._dependency_updates[updating_object] = update_id
      self._notify_observers(update_id=update_id)
```

# Proposed implementation
In setting up the dependent parameter, in the `make_dependent_on` method, before any of the parameters values are changed, a ping is sent to all its observers (if any) by calling the `_ping_observers` method (implemented in the `DescriptorNumber`):
```python
def _ping_observers(self, ping_origin=None) -> None:
    """Ping all observers to check if any cyclic dependencies has been introduced.

    :param ping_origin: Unique_name of the origin of this ping. Used to avoid cyclic depenencies.
    """
    if ping_origin == self.unique_name:
        raise RuntimeError('\n Cyclic dependency detected!\n' +
                f'An update of {self.unique_name} leads to it updating itself.\n' +
                'Please check your dependencies.')
    if ping_origin is None:
        ping_origin = self.unique_name
    for observer in self._observers:
        observer._ping_observers(ping_origin=ping_origin)
```
Only in case this dependency-traversal doesn't raise an error, is the dependent parameter updated with the dependency expression.
This approach handles the edge-case described in #10, it simplifies the code base by making the `_notify_observers` and `_update` methods simpler, and it get rids of two attributes: the `update_id_iterator` in the `global_object` and the `_dependency_updates` dictionary on Parameter, since the unique_name of the Parameter issuing the ping is now used instead.</div>

damskii9992:


rozyczko:
Currently, `legacy` has no `__init__.py`, meaning that one can't import from its content. Please consider adding the file so the legacy classes are accessible.

src/easyscience/__init__.py
```diff
@@ -1,31 +1,24 @@
...
-
-
-# alias for global_object, remove later
-def __getattr__(name):
-    if name == 'borg':
-        warnings.warn(
-            "The 'borg' has been renamed to 'global_object', this alias will be deprecated in the future", DeprecationWarning
-        )  # noqa: E501
-        print("The 'borg' has been renamed to 'global_object', this alias will be deprecated in the future")
-        return global_object
```
rozyczko, line 31: There goes our `borg`. With barely a whimper.

src/easyscience/base_classes/based_base.py
```diff
@@ -5,31 +5,27 @@
...
 from easyscience import global_object
-from easyscience.Utils.classTools import addLoggedProp
 
-from .core import ComponentSerializer
-from .variable import Parameter
-from .variable.descriptor_base import DescriptorBase
+from ..io import SerializerComponent
+from ..variable import Parameter
 
 if TYPE_CHECKING:
```
rozyczko, line 18: This is actually a nice trick we can use elsewhere to assure no circular dependencies on imports, when only doing type checks. The issue popped up a few times recently in other modules.

src/easyscience/base_classes/based_base.py
```diff
@@ -155,7 +151,7 @@ def get_parameters(self) -> List[Parameter]:
                 par_list.append(item)
         return par_list
 
-    def _get_linkable_attributes(self) -> List[V]:
+    def _get_linkable_attributes(self) -> List[DescriptorBase]:
```
rozyczko, line 154: Removing `TypeVar` and replacing it with just the base class _can_ introduce subtle issues.
The return value will be marked as `DescriptorBase` even if the passed in type is a subclass of `DescriptorBase`.
Subsequent type checks might fail (if e.g. `item_list` elements are then queried for methods which only exist on the derived class).
This is very likely moot, though, as we don't do those checks anyway. Just something to keep in mind.
`TypeVar`-ing the checks is a good practice in base classes, when we want the actual type to propagate.

src/easyscience/fitting/calculators/interface_factory.py
```diff
@@ -3,34 +3,28 @@
...
     """
 
-    def __init__(self, interface_list: List[_C], *args, **kwargs):
-        self._interfaces: List[_C] = interface_list
-        self._current_interface: _C
-        self.__interface_obj: _M = None
+    def __init__(self, interface_list: List[ABCMeta], *args, **kwargs):
+        self._interfaces: List[ABCMeta] = interface_list
+        self._current_interface: ABCMeta
+        self.__interface_obj: None = None
```
rozyczko, line 27: This is a very bizarre assignment. Both the original and this.
It's probably better/safer to use `Any` here

src/easyscience/fitting/calculators/interface_factory.py
```diff
@@ -174,7 +168,7 @@ def generate_bindings(self, model, *args, ifun=None, **kwargs):
                 prop._callback = item.make_prop(item_key)
                 prop._callback.fset(prop_value)
 
-    def __call__(self, *args, **kwargs) -> _M:
+    def __call__(self, *args, **kwargs) -> None:
```
rozyczko, line 171: Usage of `_M` in this class makes NO SENSE.
This really should be
```python
_M = TypeVar('_M')

class InterfaceTemplate(Generic[_M]):
    def __init__(self):
        self.__interface_obj: Optional[_M] = None

    def __call__(self, *args, **kwargs) -> Optional[_M]:
        return self.__interface_obj
```
if we want the class to remember what `_M` is, or even better, skip `Optional` and assign later

```python
class InterfaceTemplate(Generic[_M]):
    def __init__(self, obj: _M):
        self.__interface_obj: _M = obj

    def __call__(self, *args, **kwargs) -> _M:
        return self.__interface_obj
```

rozyczko: (mentioned because you replaced TypeVar with explicit types)

src/easyscience/global_object/global_object.py
```diff
@@ -2,11 +2,7 @@
 #  SPDX-License-Identifier: BSD-3-Clause
 #  © 2021-2023 Contributors to the EasyScience project <https://github.com/easyScience/EasyScience
 
-__author__ = 'github.com/wardsimon'
-__version__ = '0.1.0'
```
rozyczko, line 6: Maybe add your name to the file, alongside Simon's?
Removal of versioning is a good idea.

src/easyscience/io/serializer_base.py
```diff
@@ -240,9 +235,9 @@ def _convert_from_dict(d):
             if '@module' in d and '@class' in d:
                 modname = d['@module']
                 classname = d['@class']
-                # if classname in DictSerializer.REDIRECT.get(modname, {}):
-                #     modname = DictSerializer.REDIRECT[modname][classname]["@module"]
-                #     classname = DictSerializer.REDIRECT[modname][classname]["@class"]
+                # if classname in SerializerDict.REDIRECT.get(modname, {}):
+                #     modname = SerializerDict.REDIRECT[modname][classname]["@module"]
+                #     classname = SerializerDict.REDIRECT[modname][classname]["@class"]
```
rozyczko: delete commented out code

src/easyscience/io/serializer_base.py
```diff
@@ -257,46 +252,33 @@ def _convert_from_dict(d):
...
+            encoder = SerializerBase()
+        T_ = type(obj)
+        if issubclass(T_, (list, tuple, MutableSequence)):
+            # Is it a core MutableSequence?
+            if hasattr(obj, 'encode') and obj.__class__.__module__ != 'builtins':  # strings have encode
+                return encoder._convert_to_dict(obj, skip, full_encode, **kwargs)
+            else:
+                return [self._recursive_encoder(it, skip, encoder, full_encode, **kwargs) for it in obj]
+        if isinstance(obj, dict):
+            return {kk: self._recursive_encoder(vv, skip, encoder, full_encode, **kwargs) for kk, vv in obj.items()}
```
rozyczko, line 277: Is there a reason to make this a class method? It doesn't use `self` anywhere, so can be either moved out or at least made static

src/easyscience/io/serializer_component.py
```diff
@@ -0,0 +1,81 @@
...
+class SerializerComponent:
+    """
+    This base class adds the capability of saving and loading (encoding/decoding, serializing/deserializing) easyscience 
+    objects via the `encode` and `decode` methods. 
+    The default encoder is `SerializerDict`, which converts the object to a dictionary.
+
+    Shortcuts for dictionary and encoding is also present.
+    """
+
+    _CORE = True
```
rozyczko: probably out of scope, since you just moved the file, but this isn't used anywhere so likely can be removed

rozyczko (changes requested):
Additionally, since you're touching most of the files - consider updating the copyright year in the file headers to 2025.

src/easyscience/__init__.py
damskii9992: Yup. No more backwards compatibility :D 
It was time.

src/easyscience/base_classes/based_base.py
damskii9992: Yeah, I was actually looking into what it was because I wanted to remove it, but upon finding out what it did I was pleasantly surprised. We should try to remember to use this when we import classes just for type-checks in the future :)

src/easyscience/global_object/global_object.py
damskii9992: I don't think we should keep author names in all our files headers . . .
We already have our names on the published pypi package. That should be enough. Or do you not think so?

damskii9992:
> Currently, `legacy` has no `__init__.py`, meaning that one can't import from its content. Please consider adding the file so the legacy classes are accessible.

The legacy classes are not working classes, since they are not updated/maintained to follow new syntax, as such making them importable doesn't make sense. This directory is basically a trash bin which we will empty later on.

src/easyscience/io/serializer_base.py
damskii9992: Hmm, good point. We probably have a few utility class methods scattered around which could be static methods. I tried changing this one and no tests failed. Which is a little concerning, considering the way to call the method changes and thus stuff should've broken . . .

damskii9992: But is there a reason to make it a static method? The only reason I can imagine is if you want to use the method from outside the class somewhere . . .

damskii9992:
Hmm, regarding your comments on type hints, how serious are we about type hinting? Because I just tried running a type hinter on EasyScience and it doesn't just yell at me, it screams at me in guttural satanic.

PR #128, opened 2025-06-16
Add test that fixed parameters do not change

henrikjacobsenfys:
It was discovered that this test was missing.


damskii9992 (approved):
Looks good to me

PR #130, opened 2025-07-09
Improve descriptor repr and pin bumps

### LINKED ISSUE #59: Improve presentation of very small and very large Parameters/DescriptorNumbers
We should add a simple check: if a Parameter/DescriptorNumber is larger than e.g. 1e5, or smaller than 1e-3, we represent it as `x.xxxx*10^yy` - otherwise it becomes hard to read:
example

We can consider smarter solutions that also check the variance, so instead of e.g. `10.0000 +-1.0000`, we write `10.0 +-1.0`.

This is not a high priority at all.

henrikjacobsenfys:
Closes #59

### DIFF
--- a/src/easyscience/variable/descriptor_number.py
+++ b/src/easyscience/variable/descriptor_number.py
@@ -299,7 +299,7 @@ class DescriptorNumber(DescriptorBase):
         string = '<'
         string += self.__class__.__name__ + ' '
         string += f"'{self._name}': "
-        if np.abs(self._scalar.value)>1e4 or (np.abs(self._scalar.value)<1e-4 and np.abs(self._scalar.value)>0):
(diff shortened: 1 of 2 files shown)


src/easyscience/variable/descriptor_number.py
```diff
@@ -299,9 +299,15 @@ def __repr__(self) -> str:
         string = '<'
         string += self.__class__.__name__ + ' '
         string += f"'{self._name}': "
-        string += f'{self._scalar.value:.4f}'
-        if self.variance:
-            string += f' \u00b1 {self.error:.4f}'
+        if np.abs(self._scalar.value)>1e4 or (np.abs(self._scalar.value)<1e-4 and np.abs(self._scalar.value)>0):
```
damskii9992: You can get rid of the 2nd np.abs and just check if _scalar.value == 0.

damskii9992 (approved):

PR #132, opened 2025-09-02
Case fixes

rozyczko:
Important updates for the module to work with crystallography and reflectometry.
We need to unpin `bumps` for refnx to install properly. This means updating the API, which changed between 1.x and 2.x

### DIFF
--- a/src/easyscience/base_classes/based_base.py
+++ b/src/easyscience/base_classes/based_base.py
@@ -158,6 +158,7 @@ class BasedBase(SerializerComponent):
         :return: List of `Descriptor`/`Parameter` objects.
         """
         item_list = []
+        from ..variable.descriptor_base import DescriptorBase
         for key, item in self._kwargs.items():
             if hasattr(item, '_get_linkable_attributes'):
                 item_list = [*item_list, *item._get_linkable_attributes()]
--- a/src/easyscience/base_classes/based_base.py
+++ b/src/easyscience/base_classes/based_base.py
@@ -14,10 +14,10 @@ from easyscience import global_object
 
 from ..io import SerializerComponent
 from ..variable import Parameter
+from ..variable.descriptor_base import DescriptorBase
 
 if TYPE_CHECKING:
     from ..fitting.calculators import InterfaceFactoryTemplate
-    from ..variable.descriptor_base import DescriptorBase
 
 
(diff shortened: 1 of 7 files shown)


src/easyscience/base_classes/based_base.py
```diff
@@ -158,6 +158,7 @@ def _get_linkable_attributes(self) -> List[DescriptorBase]:
         :return: List of `Descriptor`/`Parameter` objects.
         """
         item_list = []
+        from ..variable.descriptor_base import DescriptorBase
```
damskii9992: I would remove this line and move the import from the top of the file out of the `if TYPE_CHECKING`.
We shouldn't be importing within methods.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -63,7 +63,7 @@ def all_methods() -> List[str]:
     @staticmethod
     def supported_methods() -> List[str]:
         # only a small subset
-        methods = ['scipy.leastsq', 'amoeba', 'newton', 'lm']
```
damskii9992, line 66: `scipy.leastsq` was replaced with MPFit, shouldn't we just replace this line with 'mpfit'?

damskii9992 (changes requested):

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Actually, it isn't immediately available through our interface

```python
>>> import bumps
>>> bumps.__version__
'1.0.2'

>>> from bumps.fitters import FIT_AVAILABLE_IDS
>>> FIT_AVAILABLE_IDS
['amoeba', 'de', 'dream', 'newton', 'lm', 'pt']


Adding it would be a proper task (validation, tests etc)
```

damskii9992: Ahh, I see. Well, it was probably meant to be added eventually, hence its appearance in this list. So maybe just update it (or add an issue about it) and we can return to it at a later point?

rozyczko: https://github.com/easyscience/corelib/issues/134

PR #136, opened 2025-09-18
New _clear

rozyczko:
Old `_clear` wasn't very performing.

Quick analysis and profiling helped improve its performance significantly. This will be especially useful for parallelization efforts.

The added `tests\performance` checks show that the chosen value for garbage control threshold is a good average for a variety of store sizes.


rozyczko:
Closing as not a necessary issue atm

PR #137, opened 2025-09-18
Changed add_vertex behaviour on identical unique name

rozyczko:
We throw on creating a new map vertex if the parameter has the same unique name.
This makes sense when a typical job is being run.
However, in a multiprocessing environment, we can't guarantee this behaviour, especially considering the `global_object` is a singleton and will be the same across the CPU pool.
One solution would be to allow multiple copies of `global_object` in MPI, but this seems a complex and error prone (at least at the moment). The alternative is to modify the unique name if it clashes with one in the map.

This is what I am proposing here.

### DIFF
--- a/src/easyscience/global_object/map.py
+++ b/src/easyscience/global_object/map.py
@@ -126,7 +126,19 @@ class Map:
     def add_vertex(self, obj: object, obj_type: str = None):
         name = obj.unique_name
         if name in self._store.keys():
-            raise ValueError(f'Object name {name} already exists in the graph.')
+            # modify the name to make it unique:
+            #  add "_1" or increment the number if already exists
+            suffix = 1
+            base_name = name
+            new_name = base_name + '_' + str(suffix)
+            while new_name in self._store.keys():
+                suffix += 1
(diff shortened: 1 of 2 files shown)


henrikjacobsenfys:
It seems like this risks renaming unique names which may be problematic for dependent Parameters? I.e. imagine I've made a Parameter a somewhere:
`a=Parameter(name='a',unique_name='par1',value=1)

And then later (perhaps in a different file) want to make a Parameter c which depends on b
```
b=Parameter(name='b',unique_name='par1',value=2)
c=Parameter(name='c',value=1)
c.make_dependent_on('2*"par1"')
```

Previously, this would throw an error, and I would know that I had messed up. Now it changes the unique name of b to `par1_1` and makes c dependent on a instead. Unless I know exactly how c should behave, I'm in trouble. Granted, it's my own fault for 1) setting the unique_name explicitly and 2) not realising I've used the same unique_name twice, but wanted to at least mention it.

rozyczko:
I'll close this PR as it got overriden by the proper fix in #175

PR #138, opened 2025-09-19
Dict size changed bug

rozyczko:
Fixed the `RuntimeError: dictionary changed size during iteration` issue in the `vertices` method by implementing thread-safety: iterations handling modifications during garbage collection.
This seems to have been a concurrency problem with weak references, where the dictionary is being modified during iteration.


rozyczko:
Nope. More work needed.

Take the attached script (.txt -> .py cause GitHub)
[run_test_suite.txt](https://github.com/user-attachments/files/22681211/run_test_suite.txt)
Run it in say, `EasyReflectometryLib` and see that it fails for `master` branch but it also fails for this branch (albeit fewer failures on average).

PR #140, opened 2025-09-19
Issue133 fitting weights

### LINKED ISSUE #133: Default fitting weights
When the user does not provide the fitter with weights to each data-point, we currently simply set the weights according to:
```python
if weights is None:
    weights = np.sqrt(np.abs(y))
```
i.e. we set the weights depending on the objective values. This is an issue for 2 reasons, the first of which is what prompted me to notice this issue:

1. When the objective y-values are 0, the weights become 0 and since we divide by weights we get `Inf`/ `NaN` errors.
2. If the objective y-values varies greatly, such as in a diffraction spectrum, the peaks would be valued infinitely higher than any other points in the spectrum.

I propose we set the default weights to 1 across the different minimizers. I.e. when users don't provide a weight, we simply weigh all data-points equally.

damskii9992:
Small PR to fix issue #133

### DIFF
--- a/src/easyscience/fitting/minimizers/minimizer_bumps.py
+++ b/src/easyscience/fitting/minimizers/minimizer_bumps.py
@@ -101,9 +101,13 @@ class Bumps(MinimizerBase):
         """
         method_dict = self._get_method_kwargs(method)
 
-        weights = np.asarray(weights)
-        if weights.ndim > 1:
-            raise ValueError('Weights must be a 1D array.')
+        x, y, weights = np.asarray(x), np.asarray(y), np.asarray(weights)
+
+        if y.shape != x.shape:
+            raise ValueError('x and y must have the same shape.')
+        
+        if weights.shape != x.shape:
+            raise ValueError('Weights must have the same shape as x and y.')
 
         if not np.isfinite(weights).all():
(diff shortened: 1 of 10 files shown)


src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -101,8 +101,15 @@ def fit(
...
-            weights = np.sqrt(np.abs(y))
+        weights = np.asarray(weights)
+        if weights.ndim > 1:
+            raise ValueError('Weights must be a 1D array.')
+
+        if not np.isfinite(weights).all():
+            raise ValueError('Weights cannot be NaN or infinite.')
+        
+        if (weights <= 0).any():
+            raise ValueError('Weights must be strictly positive and non-zero.')
```
rozyczko, line 116: maybe it would be prudent to also check the size, which needs to be the same as `x`

rozyczko (approved):
LGTM, the only issue is to potentially add another check

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: There was also no check on y. I added checks to both and updated the unit tests.

Issue #143, opened 2025-09-29
full_value assignment fails when unit present


rozyczko:
PROBLEM:
- UnitError when accessing Parameter.full_value after interface assignment
- Error message: "Expected unit Å, got dimensionless"
- Occurred when accessing: any parameter in any library after redoing bindings of the interface
- Also affected arithmetic operations between parameters

ROOT CAUSE:
- When model.interface = calculator is assigned, generate_bindings() sets up callbacks
- The callback function returns raw numpy.float64 values (dimensionless)
- Parameter.full_value property tries to compare dimensionless callback value with scipp.Variable that has units
- This comparison: scalar != self._scalar triggered UnitError

PR #144, opened 2025-09-29
fix for #143 full_value assignment fails when unit present

rozyczko:
Enhanced Parameter.full_value property:

1. Check if callback returns a scipp Variable or raw number
2. If raw number, convert to scipp scalar with original unit
3. If dimensionless scipp Variable but original has unit, apply original unit
4. Handle UnitError with fallback to preserve original unit and update value only

Also, added a test

### DIFF
--- a/src/easyscience/variable/parameter.py
+++ b/src/easyscience/variable/parameter.py
@@ -314,8 +314,24 @@ class Parameter(DescriptorNumber):
         """
         if self._callback.fget is not None:
             scalar = self._callback.fget()
-            if scalar != self._scalar:
-                self._scalar = scalar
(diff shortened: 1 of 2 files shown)


henrikjacobsenfys:
How do we know the unit of the callback when it's a raw number? Or more precisely, if it's a raw number, is there ever a situation where the unit of the callback is different from the one in `self._scalar`? I.e. `self._scalar` is in Å and the callback gives a numpy.float64 with implied unit nm? (Just trying to think of any way this might break, and this is the only one I could think of - the fix looks good to me)

rozyczko:
closed due to a better fix in https://github.com/easyscience/corelib/pull/145

PR #147, opened 2025-10-16
Macos 15 intel

rozyczko:
Added explicit `macos-15-intel` directive following scipp.


damskii9992 (approved):
Let's just merge this in :)

PR #148, opened 2025-10-16
Constraints serialization

### LINKED ISSUE #149: Dependent parameters serialization
We need to be able to save/retrieve parameters with dependencies.

rozyczko:
Details in the `PARAMETER_DEPENDENCY_SERIALIZATION.md` file.
The file is to be moved to the docs once they're actually present.


src/easyscience/variable/parameter_dependency_resolver.py
```diff
@@ -0,0 +1,112 @@
...
+            parameters.append(item)
+        elif isinstance(item, dict):
+            for value in item.values():
+                _collect_parameters(value, parameters)
+        elif isinstance(item, (list, tuple)):
+            for element in item:
+                _collect_parameters(element, parameters)
+        elif hasattr(item, '__dict__'):
+            # Check object attributes
+            for attr_name in dir(item):
```
seventil: afaik, properties will not be present in any object's `__dict__`, but in it's class' `__dict__` and you can check if class' attribute is a property, if I remember correctly, isinstance(c.attr, property) (To make sure it's only parameters, that are taken from the class' `__dict__`. 
Same in `get_parameters_with_pending_dependencies`

src/easyscience/variable/parameter_dependency_resolver.py
```diff
@@ -0,0 +1,112 @@
...
+
+from __future__ import annotations
+
+from typing import Any
+from typing import List
+
+from .parameter import Parameter
+
+
+def resolve_all_parameter_dependencies(obj: Any) -> None:
```
seventil, line 14: Might make sense to make a function in a similar fasion, but on a higher level: to take in a collection/object with serialized parameters and return the same type of a dict/list/tuple/obj with Parameters deserialized and with ensured resolved parameter dependencies (basically an umbrella for)
```
        new_params = {}
        for name, data in params_data.items():
            new_params[name] = Parameter.from_dict(data)
        resolve_all_parameter_dependencies(new_params)
```

rozyczko: good idea, thanks!

src/easyscience/variable/parameter.py
```diff
@@ -550,6 +595,68 @@ def _process_dependency_unique_names(self, dependency_expression: str):
...
+            for key, dep_obj in self._dependency_map.items():
+                # Store dependency_id if available (more reliable)
+                if hasattr(dep_obj, '__dependency_id'):
+                    d['_dependency_map_dependency_ids'][key] = dep_obj.__dependency_id
+                # Also store unique_name as fallback
+                if hasattr(dep_obj, 'unique_name'):
+                    d['_dependency_map_unique_names'][key] = dep_obj.unique_name
+                else:
+                    # Last resort fallback for objects without unique_name
+                    d['_dependency_map_unique_names'][key] = str(dep_obj)
```
henrikjacobsenfys: Would it perhaps be better to raise an error if the object doesn't have a unique name? A far as I know, every object should have a unique name, so if it doesn't, something has gone very wrong.

src/easyscience/variable/parameter_dependency_resolver.py
rozyczko: A very good idea. Implemented

src/easyscience/variable/parameter_dependency_resolver.py
rozyczko: Another good comment! Modified to query the class attribute instead.

PARAMETER_DEPENDENCY_SERIALIZATION.md
```diff
@@ -0,0 +1,216 @@
+# Parameter Dependency Serialization
+
+This document explains how to serialize and deserialize `Parameter` objects that have dependencies.
+
+## Overview
+
+Parameters with dependencies can now be serialized to dictionaries (and JSON) while preserving their dependency relationships. After deserialization, the dependencies are automatically reconstructed using the `unique_name` attribute to match parameters.
```
henrikjacobsenfys: And the dependency_id

henrikjacobsenfys (commented):
So this is essentially bringing back the unique ID's as far as I understand? If this is needed to make serialization work, then it looks good to me.

PARAMETER_DEPENDENCY_SERIALIZATION.md
rozyczko: yup! forgot to update.

src/easyscience/variable/parameter.py
rozyczko: Yes, lack of unique_name should make the loader fail earlier already (hopefully). If there's no uniqe_name at this point, we should definitely throw.

src/easyscience/variable/parameter_dependency_resolver.py
```diff
@@ -0,0 +1,147 @@
...
+                    class_attr = getattr(type(item), attr_name, None)
+                    if isinstance(class_attr, property):
+                        try:
+                            attr_value = getattr(item, attr_name)
+                            _collect_parameters(attr_value, parameters)
+                        except (AttributeError, Exception):
+                            # log the exception
+                            print(f"Error accessing property '{attr_name}' of {item}")
+                            # Skip properties that can't be accessed
+                            continue
```
damskii9992: Wouldn't it be much simpler to simply use the global map for this?
```python
all_parameters = [param for param in global_map.vertices() if param isinstance DescriptorNumber]
```
This does get all Parameters, and not just the ones within the given object, but I don't see why that would be a problem.

damskii9992 (commented):
Why do we need to be able to resolve dependencies specifically in an object hierarchy?
We can make all of this much simpler if we simply use the global map and always resolve all pending dependencies.
```python
all_parameters = [param for param in global_map.vertices() if param isinstance DescriptorNumber && hasattr(param, '_pending_dependency_string')]
```
For my remaining comments I made another PR with my suggested changes. It is mostly integrating your code further with the observer pattern and getting rid of the unique_name fallback as I don't see any use of that.

PR #151, opened 2025-10-16
Interface factory unit tests

rozyczko:
This class had ZERO unit test coverage. Now we have it covered.

Also, added codecov.

For this branch, check out
https://app.codecov.io/gh/EasyScience/corelib/tree/interface_factory_unit_tests


rozyczko:
The current unit test coverage is a bit over 75%. Need to improve that to at least 80%

rozyczko:
It's 79.71% 💀 
I'm stopping here

tests/unit_tests/variable/test_parameter.py
```diff
@@ -635,14 +635,16 @@ def test_set_value_dependent_parameter(self, normal_parameter: Parameter):
         with pytest.raises(AttributeError):
             dependent_parameter.value = 3
 
+    @pytest.mark.skip(reason="No longer relevant")
```
henrikjacobsenfys: Should we delete the tests that are not relevant?

tests/unit_tests/global_object/test_undo_redo_comprehensive.py
```diff
@@ -0,0 +1,662 @@
...
+        """Test undo/redo for creation"""
+        nd = NotarizedDict()
+        # Actually create the key first so DictStack can set it up correctly
+        nd['key'] = 'old_value'
+        cmd = DictStack(nd, 'key', 'value')
+        
+        # After creation, undo should revert to old value or delete
+        cmd.undo()
+        # Should either be deleted or reverted
+
```
henrikjacobsenfys: Should there be an assert here?

henrikjacobsenfys (commented):
I'm not too familiar with the interface factory and global object, but it looks good to me. I have only two minor comments/questions.

tests/unit_tests/variable/test_parameter.py
rozyczko: We probably should.. I need to look again at the test to see if anything can be reused.

tests/unit_tests/global_object/test_undo_redo_comprehensive.py
rozyczko: Yes, there should be. I modified the test and added another one which I found to be useful.

damskii9992:
Shouldn't this PR go into develop? Why is it going into Master?

rozyczko:
> Shouldn't this PR go into develop? Why is it going into Master?

because I'm a blind numpty. Fixed.

rozyczko:
80.23%. There we go.

PR #152, opened 2025-10-17
Documentation

rozyczko:
Initial documentation attempt. It now properly generates and updates `gh-pages` branch https://github.com/easyscience/corelib/tree/gh-pages

### DIFF
--- a/docs/src/fitting/introduction.rst
+++ b/docs/src/fitting/introduction.rst
@@ -131,13 +131,50 @@ Parameters can depend on other parameters through expressions:
     area = Parameter.from_dependency(
         name="area",
         dependency_expression="length * width",
-        dependency_map={"length": length, "width": width},
-        unit="m^2"
+        dependency_map={"length": length, "width": width}
     )
     
     # When length or width changes, area updates automatically
     length.value = 15.0
     print(area.value)  # Will be 75.0 (15 * 5)
+
+Using make_dependent_on() Method
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can also make an existing parameter dependent on other parameters using the ``make_dependent_on()`` method. This is useful when you want to convert an independent parameter into a dependent one:
+
+.. code-block:: python
+
+    # Create independent parameters
+    radius = Parameter("radius", 5.0, unit="m", min=1, max=20)
+    height = Parameter("height", 10.0, unit="m", min=1, max=50)
+    volume = Parameter("volume", 100.0, unit="m³")  # Initially independent
+    pi = Parameter("pi", 3.14159, fixed=True)  # Constant parameter
+
+    # Make volume dependent on radius and height
+    volume.make_dependent_on(
+        dependency_expression="pi * radius**2 * height",
+        dependency_map={"radius": radius, "height": height, "pi": pi}
+    )
+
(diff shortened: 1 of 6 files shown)


docs/src/fitting/introduction.rst
```diff
@@ -2,3 +2,324 @@
...
+    # Create independent parameters
+    length = Parameter("length", 10.0, unit="m", min=1, max=100)
+    width = Parameter("width", 5.0, unit="m", min=1, max=50)
+    
+    # Create dependent parameter
+    area = Parameter.from_dependency(
+        name="area",
+        dependency_expression="length * width",
+        dependency_map={"length": length, "width": width},
+        unit="m^2"
```
henrikjacobsenfys: I think this will give an error, since the unit is given by the units of length and width

docs/src/fitting/introduction.rst
```diff
@@ -2,3 +2,324 @@
...
+    )
+    
+    # When length or width changes, area updates automatically
+    length.value = 15.0
+    print(area.value)  # Will be 75.0 (15 * 5)
+
```
henrikjacobsenfys, line 140: We should also add how to make already existing `Parameter`s dependent, e.g.

```
# Make width dependent on length
width.make_dependent_on(
    "length/2",
    {"length": length})
print(width.value)  # Will be 7.5 (15 / 2)
```

docs/src/fitting/introduction.rst
```diff
@@ -2,3 +2,324 @@
...
+    from easyscience import global_object
+    
+    @pytest.fixture
+    def clear_global_map():
+        """Clear global map before each test"""
+        global_object.map._clear()
+        yield
+        global_object.map._clear()
+    
+    def test_model_fitting(clear_global_map):
```
henrikjacobsenfys, line 344: Do we also want to ask for the WHEN THEN EXPECT comments?

docs/src/getting-started/overview.rst
```diff
@@ -3,4 +3,223 @@
...
+    # Independent parameters
+    radius = Parameter("radius", 5.0, unit="m", min=0, max=100)
+    height = Parameter("height", 10.0, unit="m", min=0, max=200)
+    
+    # Dependent parameter using mathematical expression
+    volume = Parameter.from_dependency(
+        name="volume",
+        dependency_expression="3.14159 * radius**2 * height",
+        dependency_map={"radius": radius, "height": height},
+        unit="m^3"
```
henrikjacobsenfys: same as before

henrikjacobsenfys (commented):
A few minor things :)

docs/src/fitting/introduction.rst
rozyczko: indeed this is correct. `unit` is redundant here. I added it after testing to "assure" proper unit...

docs/src/fitting/introduction.rst
rozyczko: errr yes. This actually needs a separate chapter, which I conveniently forgot about :)

docs/src/fitting/introduction.rst
rozyczko: it's mostly for us. Andreas introduced this syntax, and while it is very useful in understanding the tests it is by no means a standard.

PR #154, opened 2025-10-17
Pixi support

rozyczko:
Let's start using modern tools. 
This PR standardizes build tooling to pixi for all tasks. See README.md


pixi.toml
```diff
@@ -0,0 +1,76 @@
+[project]
+name = "easyscience"
+channels = ["conda-forge", "pypi"]
```
AndrewSazonov:
```suggestion
channels = ["conda-forge"]  # conda channels only
```

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+
+[pypi-dependencies]
+# Core runtime dependencies
+asteval = "*"
+bumps = "*"
+dfo-ls = "*"
+lmfit = "*"
+numpy = "*"
+uncertainties = "*"
+scipp = "*"
```
AndrewSazonov:
```suggestion
```

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+[feature.dev.pypi-dependencies]
+# Development dependencies
+codecov = "*"
+flake8 = "*"
+matplotlib = "*"
+pytest = "*"
+pytest-cov = "*"
+ruff = "*"
+tox = "*"
+tox-gh-actions = "*"
```
AndrewSazonov:
```suggestion
easyscience = { path = ".", extras = ["dev"], editable = true }
```

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+
+[feature.docs.pypi-dependencies]
+# Documentation dependencies
+doc8 = "*"
+readme-renderer = "*"
+sphinx_autodoc_typehints = "*"
+sphinx_book_theme = "*"
+sphinx_gallery = "*"
+toml = "*"
+sphinx = "*"
```
AndrewSazonov:
```suggestion
easyscience = { path = ".", extras = ["docs"], editable = true }
```

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+clean = "rm -rf dist/ build/ *.egg-info/"
+
+# Tox tasks
+tox = "tox"
+
+# Documentation tasks
+docs-build = "sphinx-build -b html docs/src docs/_build/html"
+docs-clean = "rm -rf docs/_build/"
+
+[pypi-dependencies]
```
AndrewSazonov:
```suggestion
[pypi-dependencies] # == [feature.default.pypi-dependencies]
```

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+platforms = ["linux-64", "osx-64", "osx-arm64", "win-64"]
+
+[dependencies]
+python = ">=3.11,<3.14"
+
+[feature.build.pypi-dependencies]
+# Build dependencies
+hatchling = "<=1.21.0"
+setuptools-git-versioning = "*"
+build = "*"
```
AndrewSazonov: Consider adding a separate extra (e.g. "build") in pyproject.toml, similar to existing "dev" or "docs".
Then it can be installed like this:
```python
easyscience = { path = ".", extras = ["build"], editable = true }
```

You might also consider keeping only two Pixi environments:
* default (for non-dev usage)
* dev (including development, build, and docs dependencies)

In CI, you could then always use the dev environment. Although not every CI job will need all components of dev, dependency installation is very fast, so the performance impact is minimal. This approach simplifies maintenance by avoiding the need to remember which environment to activate in each workflow configuration.

.github/workflows/documentation-build.yml
```diff
@@ -26,22 +26,25 @@ jobs:
...
-        pip install . '.[dev,docs]'
+        run-install: false
+        cache: false
+        post-cleanup: false
+        
+    - name: Install dependencies
+      run: pixi install --environment docs
+      
+    - name: Build documentation
+      run: pixi run --environment docs docs-build
```
AndrewSazonov:
```suggestion
        environments: docs
        activate-environment: docs
        run-install: true
        cache: false
        post-cleanup: false

    - name: Build documentation
      run: pixi run docs-build
```

AndrewSazonov: You can install and activate the docs environment directly during the Pixi setup step, and this can be applied to other actions too.

README.md
```diff
@@ -19,6 +19,47 @@ Or direct from the repository:
 
 ```pip install https://github.com/easyScience/EasyScience```
 
+### Development with Pixi
```
AndrewSazonov: Consider keeping the README concise and instead linking to CONTRIBUTING.md, where this information is already documented.

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+
+[environments]
+default = ["build", "dev"]
+dev = ["build", "dev"]
+docs = ["build", "docs"]
+ci = ["build", "dev", "docs"]
+
+[tasks]
+# Development tasks
+install-dev = "pip install -e '.[dev]'"
```
AndrewSazonov: This doesn’t seem necessary, as `pixi install dev` already does this.

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+doc8 = "*"
+readme-renderer = "*"
+sphinx_autodoc_typehints = "*"
+sphinx_book_theme = "*"
+sphinx_gallery = "*"
+toml = "*"
+sphinx = "*"
+
+[environments]
+default = ["build", "dev"]
```
AndrewSazonov: Can be removed if the default is assumed to be non-dev.

AndrewSazonov (commented):
Thanks for the PR. I have a few suggestions.

pixi.toml
rozyczko: What does the commented section give me here?

pixi.toml
rozyczko: Actually I had to put it back, since non-dev doesn't contain `ruff` which pixi needs in the default env.
I don't want to add `ruff` to main dependencies, since it is not needed there for users.

AndrewSazonov: Is `ruff` really needed in the default (non-dev) environment? why?

pixi.toml
rozyczko: Absolutely correct. thanks

pixi.toml
rozyczko: Not needed. What was needed was my understanding of pixi.toml

pixi.toml
```diff
@@ -0,0 +1,56 @@
...
+
+[feature.build.pypi-dependencies]
+# Build dependencies
+hatchling = "<=1.21.0"
+setuptools-git-versioning = "*"
+build = "*"
+
+[feature.dev.pypi-dependencies]
+# Development dependencies
+easyscience = { path = ".", extras = ["dev"], editable = true }
```
damskii9992, line 11: Why editable? Isn't this flag uneccesary?

pixi.toml
```diff
@@ -0,0 +1,56 @@
+[workspace]
+name = "easyscience"
+channels = ["conda-forge"]
```
damskii9992, line 3: Do we actually use any conda-forge packages? Why is this here?

damskii9992:
Seems like not all of Andrews suggestions have been implemented yet, am I right to assume that this PR isn't ready for another review yet?

rozyczko:
> Seems like not all of Andrews suggestions have been implemented yet, am I right to assume that this PR isn't ready for another review yet?

What do you mean? I addressed all the issues raised

pixi.toml
damskii9992: You haven't added the "build" extra to the pyproject.toml.

pixi.toml
damskii9992: You did not remove the install-dev task (which is unnecessary).

CONTRIBUTING.md
```diff
@@ -5,6 +5,50 @@ email, or any other method with the owners of this repository before making a ch
...
+pixi install
+```
+
+3. Run tests to verify everything is working:
+```bash
+pixi run test
+```
+
+### Development Workflow
+
```
damskii9992, line 35: You need to add a step to update the pixi.lock file. Otherwise our dependencies will never be updated when using pixi.

.github/workflows/python-package.yml
```diff
@@ -26,18 +26,19 @@ jobs:
...
-      run: |
-        pip install tox tox-gh-actions coverage
-        tox
+    
+    - name: Setup Pixi
+      uses: ./.github/actions/setup-pixi
+        
+    - name: Set Python version
+      run: pixi add python=${{ matrix.python-version }}
+
```
damskii9992, line 35: We might want a step here which update the pixi environment, otherwise we won't know if new dependencies suddenly break our workflows since pixi uses the lock file.

damskii9992 (commented):
What exactly do we gain from using pixi versus just using uv?
Pixi is a slightly complex tool so we should be clear why exactly we need to implement it.
The environment lock files produced automatically is one nice feature, but does that warrant the added complexity of introducing pixi to our development?

pixi.toml
rozyczko: It's preferred. There's no copying to site-packages and development done in source code is immediately available without reinstalling. It's the equivalent to `pip install -e`

pixi.toml
rozyczko: We need conda-forge for python. You can't do it with pip, as pip needs python.

CONTRIBUTING.md
rozyczko: added to python-ci.yml

AndrewSazonov: I think we should run `pixi update` locally only when we explicitly want to update dependencies, rather than as part of the CI workflow.

First, this ensures reproducibility. CI should use exactly the same dependencies as we have locally.
Second, it gives us control over dependency updates and lockfile changes, instead of triggering them on every GitHub run.

If we want to test against unpinned dependencies, I would suggest to create a separate nightly workflow dedicated to that purpose.

damskii9992: That makes sense. We do need the nightly workflow though, as we cannot NOT test our software against the newest versions of packages (as they are what users will get when installing our software).

.github/workflows/python-ci.yml
```diff
@@ -19,11 +19,20 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@v0.9.0
+        with:
+          environments: dev
+          activate-environment: dev
+          run-install: true
+          cache: false
+          post-cleanup: false
+
+      - name: Run linting
+        run: pixi run lint
```
damskii9992, line 25: what about the format task? Shouldn't we also run that?

.github/workflows/python-ci.yml
```diff
@@ -19,11 +19,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - uses: astral-sh/ruff-action@v3
+      - uses: prefix-dev/setup-pixi@v0.9.0
+        with:
+          environments: dev
+          activate-environment: dev
+          run-install: true
+          cache: false
```
damskii9992: Why are we setting cache: false? 
Since we're using the pixi-lock environment in every run, we might as well cache the environment and re-use it throughout the workflow. Of course this is not the case for the nightly workflow where we update the environment.

.github/workflows/python-package.yml
```diff
@@ -26,18 +26,19 @@ jobs:
...
-      run: |
-        pip install -e '.[dev]'
-        python -m build
-    - name: Test with pytest
-      run: |
-        pip install tox tox-gh-actions coverage
-        tox
+    
+    - name: Setup Pixi
+      uses: ./.github/actions/setup-pixi
```
damskii9992, line 31: Why do we use a different action to setup pixi here?

.github/workflows/python-package.yml
```diff
@@ -26,18 +26,19 @@ jobs:
...
+        
+    - name: Set Python version
+      run: pixi add python=${{ matrix.python-version }}
+      
+    - name: Build package
+      run: pixi run build
+      
+    - name: Run tests with tox
+      run: pixi run tox
```
damskii9992, line 40: Since we run tox in the python-ci workflow, can't we remove it here? I understand that being doubly sure is double as good, but it is also annoying to have to wait for this step when trying to make a release, especially since it should be unnecessary.

CONTRIBUTING.md
```diff
@@ -5,6 +5,50 @@ email, or any other method with the owners of this repository before making a ch
...
+- **Run unit tests only**: `pixi run test-unit`
+- **Check code style**: `pixi run lint`
+- **Fix code style**: `pixi run lint-fix`
+- **Format code**: `pixi run format`
+- **Build package**: `pixi run build`
+- **Build documentation**: `pixi run docs-build`
+
+### Before Submitting a Pull Request
+
+1. Ensure all tests pass: `pixi run test`
```
damskii9992: Before this step, add a "run: pixi run update-lock" since we should update the lock-file on most of our pull-requests.

pixi.toml
```diff
@@ -0,0 +1,54 @@
...
+# Code quality tasks
+lint = "ruff check ."
+lint-fix = "ruff check . --fix"
+format = "ruff format ."
+format-check = "ruff format . --check"
+
+# Build tasks
+build = "python -m build"
+clean = "rm -rf dist/ build/ *.egg-info/"
+update-lock = "pixi install"
```
damskii9992: This command should be `pixi update`, right?

pixi.toml
```diff
@@ -0,0 +1,54 @@
...
+
+[feature.build.pypi-dependencies]
+# build dependencies
+easyscience = { path = ".", extras = ["build"], editable = true }
+
+[environments]
+# Default environment with basic build setup
+default = ["build"]
+# Development environment with build, dev, and docs features
+dev = ["build", "dev", "docs"]
```
damskii9992, line 25: Since we're using the dev environment everywhere, is there even a reason to split up our dependencies into the sub-groups of documentation, build and dev?

damskii9992 (changes requested):

.github/workflows/python-package.yml
rozyczko: That's a spillover from another branch, which is to be merged with `pixi` - creating a local action to simplify pixi setup. After merging, we no longer need to add all the `with:...` lines.
So here - it's a bug, but after merging (already approved) `pixi-test`, it'll be a feature.

.github/workflows/python-ci.yml
rozyczko: This avoids potential issues with stale or corrupted caches. This can happen if dependencies change between runs (e.g., new package versions or lockfile updates). I think it promotes reliability and reproducibility, especially for testing across multiple Python versions or OSes. Cache issues can be hard to spot, causing intermittent failures that are tough to reproduce.
Once the build process is shown to be stable, I will switch to `cache:true` to optimize performance.

.github/workflows/python-package.yml
rozyczko: This assures that the package which is to be uploaded to GH is working well. Again, as with cache clearing issue - 99.9% of the time both python-ci and python-package testing will be identical, but additional unit test run before uploading the package is prudent. Note that this workflow is only run when pushed or PR-ed to `master`. Not that often.

pixi.toml
rozyczko: whoops

pixi.toml
rozyczko: I think there still is. If you want to develop the code you only need ['dev'] environment, you don't need to install sphinx and other. This speeds up creating dev environments locally.
In the CI they are merged into 'dev' as above.

.github/workflows/python-ci.yml
damskii9992: As far as I could understand from the documentation of the action, the cache only lives for the duration of the workflow, but I guess the thing is fast enough that we don't really need to cache the environments to speed it up. 👍

src/easyscience/variable/parameter.py
```diff
@@ -532,22 +572,26 @@ def _process_dependency_unique_names(self, dependency_expression: str):
...
-        inputted_unique_names += re.findall('(\".+?\")', dependency_expression)
+        inputted_unique_names = re.findall("('.+?')", dependency_expression)
+        inputted_unique_names += re.findall('(".+?")', dependency_expression)
 
         clean_dependency_string = dependency_expression
         existing_unique_names = self._global_object.map.vertices()
         # Add the unique names of the parameters to the ASTEVAL interpreter
         for name in inputted_unique_names:
-            stripped_name = name.strip("'\"")
+            stripped_name = name.strip('\'"')
```
damskii9992: Same here

pixi.toml
```diff
@@ -0,0 +1,54 @@
...
+# Development tasks
+test = "pytest --cov=src/easyscience tests --cov-branch --cov-report=xml:coverage-unit.xml"
+test-unit = "pytest tests/unit_tests"
+test-integration = "pytest tests/integration_tests"
+
+# Code quality tasks
+lint = "ruff check ."
+lint-fix = "ruff check . --fix"
+format = "ruff format ."
+format-check = "ruff format . --check"
```
damskii9992, line 36: Can we rename format -> format-fix
And format-check -> format
So that formatting and linting has a similar syntax? :) 
You can also rename the lint-commands to be similar to the format commands. I'd just like then identical 😅

.github/workflows/python-ci.yml
```diff
@@ -39,17 +52,18 @@ jobs:
...
     - uses: actions/checkout@v5
 
-    - uses: actions/setup-python@v6
+    - name: Setup Pixi
+      uses: prefix-dev/setup-pixi@v0.9.0
       with:
-        python-version: ${{ matrix.python-version }}
+        run-install: false
+        cache: false
+        post-cleanup: false
```
damskii9992, line 60: I assume this should now use our composite action?

.github/workflows/python-ci.yml
```diff
@@ -69,16 +83,14 @@ jobs:
...
+
+    - name: Build package
+      run: pixi run build
 
-    - name: Install dependencies and build
-      run: |
-        pip install -e '.[dev]'
-        python -m build
+    - name: Update lock file
+      run: pixi run update-lock
```
damskii9992: Based on the discussion with Andrew, remove this again and instead make another nightly workflow where we do run this and run tests again so that we know that our software always works with the latest dependencies.

damskii9992 (commented):
Hopefully the last comments. I think if you fix these we are ready to merge this PR :)

.github/workflows/python-ci.yml
rozyczko: You assume wrong. In this case, pixi needs to set up the particular python environment (line 63) so the pre-made composite won't work (don't remember the exact error I was getting now)

pixi.toml
rozyczko:
```
lint-check = "ruff check ."
lint = "ruff check . --fix"
format-check = "ruff format . --check"
format = "ruff format ."
```
this looks the best when actually using. Cf. CONTRIBUTING.md

.github/workflows/python-ci.yml
rozyczko: Added `nightly-check.yml`

damskii9992 (approved):
Looks good and ready to merge :)

PR #155, opened 2025-10-23
Pixi test

rozyczko:
Use local actions to simplify pixi setup

Use of `uses: prefix-dev/setup-pixi@v0.9.0` in the Code_Testing job is necessary since we need an `activate-environment:`-less specs for it here. The subsequent `pixi add python=${{ matrix.python-version }}` will create and activate the required env instead

### DIFF
--- a/.github/actions/setup-pixi/action.yml
+++ b/.github/actions/setup-pixi/action.yml
@@ -1,0 +1,38 @@
+name: 'Setup Pixi Environment'
+description: 'Sets up pixi with common configuration'
+inputs:
+  environments:
+    description: 'Pixi environments to setup'
+    required: false
+    default: 'dev'
+  activate-environment:
+    description: 'Environment to activate'
+    required: false
+    default: 'dev'
+  run-install:
+    description: 'Whether to run install'
+    required: false
+    default: 'false'
+  cache:
+    description: 'Whether to use cache'
+    required: false
+    default: 'false'
+  post-cleanup:
+    description: 'Whether to run post cleanup'
+    required: false
+    default: 'false'
+  pixi-version:
+    description: 'Version of setup-pixi action to use'
+    required: false
+    default: 'v0.9.0'
+
+runs:
+  using: 'composite'
+  steps:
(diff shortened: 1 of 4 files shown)


AndrewSazonov (commented):
Looks good to me. Using composite actions is definitely a great way to create reusable workflow components shared across multiple jobs.

Also, consider defining separate features for each Python version (and corresponding environments) directly in the pixi.toml file. This allows running tests for multiple Python versions sequentially within the same runner, which can noticeably speed up CI, especially on macOS (where the number of runners is usually quite limited) and on Windows (where dependency installation tends to be slower even with Pixi).

rozyczko:
> Also, consider defining separate features for each Python version (and corresponding environments) directly in the pixi.toml file. 

You mean something you've done in diffraction?
```
[feature.py311.dependencies]
python = '3.11.*'
```

rozyczko:
I will experiment with this approach when/if the performance hit is justifying the complexity of the solution.

damskii9992 (approved):
This workflow essentially saves 6 lines of code per time you use the action (4 times total) at the cost of an extra workflow consisting of 35 lines of code . . .
This really does not seem worth it :/
You get the added benefit that you only have to change values in 1 place rather than in 4, but with ctrl+f in visual studio code that is not really a benefit . . .
I don't really mind this, but it just seems silly to me 😅

rozyczko:
> This workflow essentially saves 6 lines of code per time you use the action (4 times total) at the cost of an extra workflow consisting of 35 lines of code . . . This really does not seem worth it :/ You get the added benefit that you only have to change values in 1 place rather than in 4, but with ctrl+f in visual studio code that is not really a benefit . . . I don't really mind this, but it just seems silly to me 😅

Well, it's really a DRY issue. Instead of updating the release version of the action in 4 places, you update it in only one.
Makes maintenance easier. VSCode ctrl-f is not really the answer.
I would go even further, claiming that if we reuse ANY action like `actions/checkout@v5` across GH workflow files, we should place them in a composite action for this purpose.

PR #156, opened 2025-10-28
Change pr label workflow and add issue workflow

damskii9992:
This PR finally adds the workflow to add labels to issues and it also updates the workflow that checks labels on pull requests.


.github/workflows/verify_issue_labels.yml
```diff
@@ -0,0 +1,28 @@
...
+          secret: ${{ github.TOKEN }}
+          prefix: "[scope]"
+          addLabel: true
+          defaultLabel: "[scope] ⚠️ label needed"
+
+      - name: Require [priority] label
+        uses: trstringer/require-label-prefix@v1
+        with:
+          secret: ${{ github.TOKEN }}
+          prefix: "[scope]"
```
rozyczko: I think you meant `[priority]` here

.github/workflows/verify_issue_labels.yml
```diff
@@ -0,0 +1,28 @@
...
+    types: [opened, labeled, unlabeled]
+
+jobs:
+  check_issue_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Require [scope] label
+        uses: trstringer/require-label-prefix@v1
+        with:
+          secret: ${{ github.TOKEN }}
```
rozyczko, line 18: This should be `${{ secrets.GITHUB_TOKEN }}`

.github/workflows/verify_issue_labels.yml
```diff
@@ -0,0 +1,28 @@
+# This workflow will be verify that all issues have at
```
rozyczko: `will verify`

.github/workflows/verify_issue_labels.yml
```diff
@@ -0,0 +1,28 @@
...
+  check_issue_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Require [scope] label
+        uses: trstringer/require-label-prefix@v1
+        with:
+          secret: ${{ github.TOKEN }}
+          prefix: "[scope]"
+          addLabel: true
+          defaultLabel: "[scope] ⚠️ label needed"
```
rozyczko, line 21: If the first step fails, the second step won't run. Consider using `continue-on-error` or `if: always()` if you want both checks to run.

.github/workflows/verify_issue_labels.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Require [priority] label
+        uses: trstringer/require-label-prefix@v1
+        with:
+          secret: ${{ github.TOKEN }}
+          prefix: "[scope]"
+          addLabel: true
+          defaultLabel: "[priority] ⚠️ label needed"
+
```
rozyczko, line 31: The workflow doesn't explicitly declare permissions. GITHUB_TOKEN gives you default permissions, but I think it's better to be explicit about what the workflow needs:
```
permissions:
  issues: write
```

rozyczko (changes requested):

.github/workflows/verify_issue_labels.yml
damskii9992: I actually don't think so. Especially not if we want to make composite actions:
https://stackoverflow.com/questions/66899758/difference-between-github-token-vs-secrets-github-token

.github/workflows/verify_issue_labels.yml
damskii9992: Not sure how it could potentially fail, but added the `if: always()` just to be prudent.

rozyczko (approved):
lgtm now

PR #158, opened 2025-11-10
Remove Unique_name fallback and integrate serialization with observer…

damskii9992:
I found it easier to make another PR to your PR rather than write down all my suggested changes.

### DIFF
--- a/src/easyscience/variable/descriptor_number.py
+++ b/src/easyscience/variable/descriptor_number.py
@@ -68,8 +68,8 @@ class DescriptorNumber(DescriptorBase):
         self._observers: List[DescriptorNumber] = []
 
         # Extract dependency_id if provided during deserialization
-        if '_dependency_id' in kwargs:
-            self._dependency_id = kwargs.pop('_dependency_id')
+        if '__dependency_id' in kwargs:
+            self.__dependency_id = kwargs.pop('__dependency_id')
 
         if not isinstance(value, numbers.Number) or isinstance(value, bool):
             raise TypeError(f'{value=} must be a number')
@@ -118,14 +118,14 @@ class DescriptorNumber(DescriptorBase):
     def _attach_observer(self, observer: DescriptorNumber) -> None:
         """Attach an observer to the descriptor."""
         self._observers.append(observer)
-        if not hasattr(self, '_dependency_id'):
-            self._dependency_id = str(uuid.uuid4())
+        if not hasattr(self, '_DescriptorNumber__dependency_id'):
+            self.__dependency_id = str(uuid.uuid4())
 
     def _detach_observer(self, observer: DescriptorNumber) -> None:
         """Detach an observer from the descriptor."""
         self._observers.remove(observer)
         if not self._observers:
-            del self._dependency_id
+            del self.__dependency_id
 
     def _notify_observers(self) -> None:
         """Notify all observers of a change."""
@@ -333,8 +333,8 @@ class DescriptorNumber(DescriptorBase):
         raw_dict['value'] = self._scalar.value
         raw_dict['unit'] = str(self._scalar.unit)
         raw_dict['variance'] = self._scalar.variance
-        if hasattr(self, '_dependency_id'):
-            raw_dict['_dependency_id'] = self._dependency_id
+        if hasattr(self, '_DescriptorNumber__dependency_id'):
+            raw_dict['__dependency_id'] = self.__dependency_id
         return raw_dict
 
     def __add__(self, other: Union[DescriptorNumber, numbers.Number]) -> DescriptorNumber:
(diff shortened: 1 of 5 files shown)


src/easyscience/variable/descriptor_number.py
```diff
@@ -50,6 +51,7 @@ def __init__(
         url: Optional[str] = None,
         display_name: Optional[str] = None,
         parent: Optional[Any] = None,
+        **kwargs: Any  # Additional keyword arguments (used for (de)serialization)
```
rozyczko, line 54: what's the reason to move `_dependency_id` logic to `descriptor_number`? It is only `Parameter`s which will use it. No need to pollute the base class.

src/easyscience/variable/parameter.py
```diff
@@ -595,63 +590,26 @@ def _process_dependency_unique_names(self, dependency_expression: str):
...
-            # Convert dependency_map to use dependency_ids (preferred) and unique_names (fallback)
-            d['_dependency_map_dependency_ids'] = {}
-            d['_dependency_map_unique_names'] = {}
-
-            for key, dep_obj in self._dependency_map.items():
-                # Store dependency_id if available (more reliable)
-                if hasattr(dep_obj, '__dependency_id'):
-                    d['_dependency_map_dependency_ids'][key] = dep_obj.__dependency_id
-                # Also store unique_name as fallback
-                if hasattr(dep_obj, 'unique_name'):
```
rozyczko, line 617: agreed with the removal of `unique_id` for dependency serialization

src/easyscience/variable/descriptor_number.py
```diff
@@ -50,6 +51,7 @@ def __init__(
...
     ):
```
rozyczko, line 55: Removing the double underscore risks allowing accidental overwrite of this attribute by child classes, effectively breaking the serialization mechanism. This is why I chose having `__dependency.id` with two underscores in the parameter class.

rozyczko (changes requested):
I don't understand why you moved the logic to the base class - apart of the Parameter, no derived class will use the dependency serialization.
Also, you are losing the attribute security by removing the extra underscore.

Any good reason for both of these?

src/easyscience/variable/descriptor_number.py
damskii9992: A `Parameter` can use a `DescriptorNumber` as one of its dependencies, therefore we actually NEED to have this logic in the `DescriptorNumber` class, if we wish to be able to serialize/deserialize all dependencies.

src/easyscience/variable/parameter.py
damskii9992: I was afraid of having this fallback mechanism would cause things to occasionally work when they should fail . . .
Or when they fail we would get the wrong error. So I think it makes more sense to only rely on one or the other, and here the `dependency_id` is more robust :)

src/easyscience/variable/descriptor_number.py
damskii9992: I don't really see how or why we would accidentally overwrite it, but I see your point. 
The `dependency_id`, once set, shouldn't ever be overwritten. So name-mangling it might make it slightly more safe/robust.

damskii9992: I've pushed a commit to name-mangle the `dependency_id` again. I'm not sure I really like it though as name-mangling makes attributes quite annoying to access and it makes the code somewhat harder to write since we have to remember when to prepend the class name to the attribute and when not to.

src/easyscience/variable/parameter.py
seventil: I agree that the fallback mechanism could lead to silent fails. And there's no expectation to use such a fallback, since dependency map dependency id doesnt fallback on the `unique name`. But on the other hand it kinda repeats the responsibility of `unique_name`: have a serializable, potentially human-readable unique identification of an object. On top of having just `name`.
Name clashes of `unique_name` have to be resolved in an execution anyway, so having a clash between dependencies relying on unique_names should not be a reason to have separate dependency_id, since it falls under the umbrella of an issue of having a name clash in general. Unless I'm missing out on another reason for having dependency_id.
Resolving name clashes in name/unique_name would lift the neccesity for dependency_id

src/easyscience/variable/descriptor_number.py
rozyczko: I don't understand this statement - can you please give me an example where the dependency serialization is necessary in `DescriptorNumber`?

damskii9992: I made a unit test for it in the latest commit, but essentially:
```python
x = DescriptorNumber(name="x", value=3.0, unit="m")
y = Parameter(name="y", value=4.0, unit="m")
z = Parameter.from_dependency(
    name="z",
    dependency_expression="x + y",
    dependency_map={"x": x, "y": y},
)
```
You can make dependencies which use DescriptorNumbers in their dependency expression.

PR #159, opened 2025-11-19
Create new easyscience base class and model_base class.

damskii9992:


src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,148 @@
...
+            raise TypeError('Display name must be a string or None')
+        self._display_name = display_name
+
+    @property
+    def _arg_spec(self) -> Set[str]:
+        """
+        This method is used by the serializer to determine which arguments are needed
+        by the constructor to deserialize the object.
+        """
+        base_cls = getattr(self, '__old_class__', self.__class__)
```
rozyczko: `__old_class__` is a part of the old design, set up in `addLoggedProp` which I think you wanted to avoid?

How about
```python
spec = getfullargspec(self.__class__.__init__)
```

`NewBase` doesn't use dynamic class creation and the `__old_class__` attribute is never set on `NewBase` instances...

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,148 @@
...
+from ..io.serializer_base import SerializerBase
+
+
+class NewBase:
+    """
+    This is the new base class for library objects.
+    It provides serialization capabilities as well as unique naming and display naming.
+"""
+
+    def __init__(self, unique_name: Optional[str] = None, display_name:  Optional[str] = None):
```
rozyczko: I think `NewBase` should only deal with unique_names. Why do you consider display_name to be a part of it? It seems too unimportant - can be set by classes which inherit `NewBase`...

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,148 @@
...
+        self._global_object = global_object
```
rozyczko, line 30: Why not have parameter discovery on the level of `NewBase`? Having `__dir__` here would make more sense

src/easyscience/io/serializer_base.py
```diff
@@ -260,6 +260,35 @@ def _convert_from_dict(d):
...
+                        if hasattr(cls_, 'from_dict'):
+                            out_dict[key] = cls_.from_dict(value)
+                        else:
+                            out_dict[key] = SerializerBase._convert_from_dict(value)
+                    else:
+                        raise ValueError(f'Class {class_name} not found in module {module_name}.')
+                else:
+                    out_dict[key] = SerializerBase._convert_from_dict(value)
+        return out_dict
+
```
rozyczko, line 332: This implementation smells. Let's refactor it so it is easier to test.

rozyczko, line 332:
```python
    @staticmethod
    def _deserialize_dict(in_dict: Dict[str, Any]) -> Dict[str, Any]:
        """
        Deserialize a dictionary using from_dict for ES objects and SerializerBase otherwise.
        This method processes constructor arguments, skipping metadata keys starting with '@'.
        
        :param in_dict: dictionary to deserialize
        :return: deserialized dictionary with constructor arguments
        """
        d = {
            key: SerializerBase._deserialize_value(value)
            for key, value in in_dict.items()
            if not key.startswith('@')
        }
        return d

    @staticmethod
    def _deserialize_value(value: Any) -> Any:
        """
        Deserialize a single value, using specialized handling for ES objects.
        
        :param value:
        :return: deserialized value
        """
        if not SerializerBase._is_serialized_easyscience_object(value):
            return SerializerBase._convert_from_dict(value)
            
        module_name = value['@module']
        class_name = value['@class']
        
        try:
            cls = SerializerBase._import_class(module_name, class_name)
            
            # Prefer from_dict() method for ES objects
            if hasattr(cls, 'from_dict'):
                return cls.from_dict(value)
            else:
                return SerializerBase._convert_from_dict(value)
                
        except (ImportError, ValueError) as e:
            # Fallback to generic deserialization if class-specific fails
            return SerializerBase._convert_from_dict(value)

    @staticmethod
    def _is_serialized_easyscience_object(value: Any) -> bool:
        """
        Check if a value represents a serialized ES object.
        
        :param value: 
        :return: True if this is a serialized ES object
        """
        return (
            isinstance(value, dict) 
            and "@module" in value 
            and value["@module"].startswith("easy") 
            and '@class' in value
        )

    @staticmethod
    def _import_class(module_name: str, class_name: str):
        """
        Import a class from a module name and class name.
        
        :param module_name: name of the module
        :param class_name: name of the class
        :return: the imported class
        :raises ImportError: if module cannot be imported
        :raises ValueError: if class is not found in module
        """
        try:
            module = __import__(module_name, globals(), locals(), [class_name], 0)
        except ImportError as e:
            raise ImportError(f'Could not import module {module_name}') from e
            
        if not hasattr(module, class_name):
            raise ValueError(f'Class {class_name} not found in module {module_name}.')
            
        return getattr(module, class_name)
```

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,112 @@
...
+from typing import Optional
+
+from ..io import SerializerBase
+from ..variable import DescriptorNumber
+from ..variable import Parameter
+from .new_base import NewBase
+
+
+class ModelBase(NewBase):
+    """
```
rozyczko, line 23: yay! clean inheritance! no `__setattr__` overrides! No forced integration with global object mapping for basic functionality!

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,112 @@
...
+    This is the base class for all model classes in EasyScience.
```
rozyczko, line 24: `ModelBase` is where we should add `display_name`, I think?

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,112 @@
...
+        super().__init__(unique_name=unique_name, display_name=display_name)
+
+    def get_all_parameters(self) -> List[DescriptorNumber]:
+        """
+        Get all `Parameters` or `DescriptorNumber` objects as a list.
+
+        :return: List of `DescriptorNumber` or `Parameter` objects.
+        """
+        params = []
+        for attr_name in dir(self):
```
rozyczko, line 52: I think there's insufficient introspection here. Should we not add
```python
            if attr_name.startswith('_'):
                continue
            attr = getattr(self, attr_name)
            if callable(attr):
                continue
```
to screen out private attrs and non-param objects?

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def get_fit_parameters(self) -> List[Parameter]:
+        """
+        Get all parameters which can be fitted as a list.
+
+        :return: List of `Parameter` objects.
+        """
+        params = []
+        for attr_name in dir(self):
+            attr = getattr(self, attr_name)
```
rozyczko:
```python
return [param for param in self.get_fit_parameters() if param.fixed]
```

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,112 @@
...
+        Get all parameters which are currently free to be fitted as a list.
+
+        :return: List of `Parameter` objects.
+        """
+        params = []
+        for attr_name in dir(self):
+            attr = getattr(self, attr_name)
+            if isinstance(attr, Parameter) and not attr.fixed and attr.independent:
+                params.append(attr)
+        return params
```
rozyczko:
```python
return [param for param in self.get_fit_parameters() if not param.fixed]
```

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,112 @@
...
+    
+    @classmethod
+    def from_dict(cls, obj_dict: Dict[str, Any]) -> None:
+        """
```
rozyczko, line 93: multiple conditionals are icky. I'd like to rewrite this to be more modular, but can't be arsed now.

rozyczko (commented):
No dynamic parameter injections!
I think this is a start of something awesome.
Next stop - the interface or rather, the calculator factory...

src/easyscience/base_classes/new_base.py
rozyczko: EDIT: I'm dumb. It *is* defined in `NewBase`. Disregard.

src/easyscience/base_classes/model_base.py
damskii9992: Yeah, I considered this approach too. It's much simpler, for the other method too. The reason I didn't choose it was for performance. This implementation only runs through the list of attributes once, your suggestion runs through the produced list again to re-filter it. The performance gain is probably so minimal it isn't worth it though :/

src/easyscience/base_classes/new_base.py
damskii9992: This is why we should probably have an ADR discussion. To discuss what goes into the ground base class (this would be `NewBase`) and what goes into the derived classes :) 
But I'm not against moving it into `ModelBase`.

src/easyscience/io/serializer_base.py
damskii9992: Agreed, this is much better. I took inspiration from the `_convert_from_dict` method, which is why it ended up not being properly factorised. 
The `_is_serialized_easyscience_object` method might be taking it too far though, being a single logical check.

src/easyscience/base_classes/model_base.py
damskii9992: The `__dir__` method implemented in `NewBase` deals with hidden attributes.
Filtering out callables could be problematic if we make a class at some point which is callable. And we barely save anything anyways, since we currently only have 2 check (`isinstance(DescriptorNumber)` and `hasattr(attr, 'get_all_parameters')`), we would just be replacing 2 checks with 1 check.

src/easyscience/base_classes/model_base.py
damskii9992: I fixed it. Using the `_is_serialized_easyscience_object` method you refactored in the `SerializerBase` class. Clearly it was useful to refactor it even though it was only a simple logical check.

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,143 @@
...
+        return names
+
+    @property
+    def unique_name(self) -> str:
+        """Get the unique name of the object."""
+        return self._unique_name
+
+    @unique_name.setter
+    def unique_name(self, new_unique_name: str):
+        """Set a new unique name for the object. The old name is still kept in the map.
```
henrikjacobsenfys, line 61: Why is the old name still kept in the map? Just curious; I haven't taken the time to even try to understand how the map works.

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,96 @@
...
+        params = []
+        for attr_name in dir(self):
+            attr = getattr(self, attr_name)
+            if isinstance(attr, DescriptorNumber):
+                params.append(attr)
+            elif hasattr(attr, 'get_all_parameters'):
+                params += attr.get_all_parameters()
+        return params
+
+    def get_fit_parameters(self) -> List[Parameter]:
```
henrikjacobsenfys: I'm pretty sure the fitter uses this method to choose which Parameters to fit, so we need to be a bit careful here

src/easyscience/base_classes/new_base.py
damskii9992: This was decided back in #16 under Andrews comment and Andreas reply. Thank god that we have ADR's and paper trails to follow now to remember these things ;)

src/easyscience/base_classes/model_base.py
damskii9992: That is exactly why I named this method like this. I wanted to name it `get_fittable_parameters`, but the `Fitter` expects this method name :/ 
I will obviously make unit tests and integration tests to check that this implementation works with the `Fitter` :)

PR #162, opened 2025-11-24
Doc fixes

rozyczko:
It seems the docs branch had docs but didn't really build them properly.
Which is really interesting, since the last documentation job on that branch ran fine...

https://github.com/easyscience/corelib/actions/runs/18569649015

This fixes things.


damskii9992 (approved):
Looks good to me

PR #163, opened 2025-11-26
New base class

damskii9992:
New pull request, because of weird and overlarge large changelog in #159 after rebasing on the newest develop branch.
This PR implements the changes discussed in ADR #161: A new EasyScience base class to replace BasedBase, ObjectBase, and SerializerComponent, and another base class inheriting from that to implement model-specific functionality.

### DIFF
--- a/src/easyscience/base_classes/model_base.py
+++ b/src/easyscience/base_classes/model_base.py
@@ -4,10 +4,12 @@ from __future__ import annotations
 #  SPDX-License-Identifier: BSD-3-Clause
 #  © 2021-2025 Contributors to the EasyScience project <https://github.com/easyScience/EasyScience
 from typing import TYPE_CHECKING
-from typing import Any
-from typing import Dict
-from typing import List
-from typing import Optional
+
+if TYPE_CHECKING:
+    from typing import Any
+    from typing import Dict
+    from typing import List
+    from typing import Optional
 
 from ..io import SerializerBase
 from ..variable import DescriptorNumber
--- a/src/easyscience/base_classes/model_base.py
+++ b/src/easyscience/base_classes/model_base.py
@@ -89,7 +89,7 @@ class ModelBase(NewBase):
         return self.get_free_parameters()
 
     @classmethod
-    def from_dict(cls, obj_dict: Dict[str, Any]) -> None:
+    def from_dict(cls, obj_dict: Dict[str, Any]) -> ModelBase:
         """
         Re-create an EasyScience object with DescriptorNumber attributes from a full encoded dictionary.
 
@@ -107,9 +107,13 @@ class ModelBase(NewBase):
                     kwargs[key] = value.value
             cls_instance = cls(**kwargs)
             for key, value in parameter_placeholder.items():
-                temp_param = getattr(cls_instance, key)
-                setattr(cls_instance, '_' + key, value)
-                cls_instance._global_object.map.prune(temp_param.unique_name)
+                try:
+                    temp_param = getattr(cls_instance, key)
+                    setattr(cls_instance, '_' + key, value)
+                    cls_instance._global_object.map.prune(temp_param.unique_name)
+                except Exception as e:
+                    raise SyntaxError(f"""Could not set parameter {key} during `from_dict` with full deserialized variable. \n'
+                            This should be fixed in the class definition. Error: {e}""") from e
             return cls_instance
         else:
             raise ValueError(f'Class name in dictionary does not match the expected class: {cls.__name__}.')
--- a/src/easyscience/base_classes/model_base.py
+++ b/src/easyscience/base_classes/model_base.py
@@ -80,7 +80,7 @@ class ModelBase(NewBase):
         :return: List of `Parameter` objects.
         """
         return [param for param in self.get_fittable_parameters() if not param.fixed]
-    
+
     def get_fit_parameters(self) -> List[Parameter]:
         """
         This is an alias for `get_free_parameters`.
--- a/src/easyscience/base_classes/model_base.py
+++ b/src/easyscience/base_classes/model_base.py
@@ -73,13 +73,6 @@ class ModelBase(NewBase):
         """
         return [param for param in self.get_all_parameters() if param.independent]
 
-    def get_fit_parameters(self) -> List[Parameter]:
-        """
-        This is an alias for `get_fittable_parameters`.
-        To be removed when fully moved to new base classes and minimizer can be changed.
-        """
-        return self.get_fittable_parameters()
-
     def get_free_parameters(self) -> List[Parameter]:
         """
         Get all parameters which are currently free to be fitted as a list.
@@ -87,6 +80,13 @@ class ModelBase(NewBase):
         :return: List of `Parameter` objects.
         """
         return [param for param in self.get_fittable_parameters() if not param.fixed]
+    
+    def get_fit_parameters(self) -> List[Parameter]:
+        """
+        This is an alias for `get_free_parameters`.
+        To be removed when fully moved to new base classes and minimizer can be changed.
+        """
+        return self.get_free_parameters()
 
     @classmethod
     def from_dict(cls, obj_dict: Dict[str, Any]) -> None:
--- a/src/easyscience/base_classes/model_base.py
+++ b/src/easyscience/base_classes/model_base.py
@@ -99,7 +99,7 @@ class ModelBase(NewBase):
         if not SerializerBase._is_serialized_easyscience_object(obj_dict):
             raise ValueError('Input must be a dictionary representing an EasyScience object.')
         if obj_dict['@class'] == cls.__name__:
-            kwargs = SerializerBase._deserialize_dict(obj_dict)
+            kwargs = SerializerBase.deserialize_dict(obj_dict)
             parameter_placeholder = {}
             for key, value in kwargs.items():
                 if isinstance(value, DescriptorNumber):
--- a/src/easyscience/base_classes/model_base.py
+++ b/src/easyscience/base_classes/model_base.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+#  SPDX-FileCopyrightText: 2025 EasyScience contributors  <core@easyscience.software>
+#  SPDX-License-Identifier: BSD-3-Clause
+#  © 2021-2025 Contributors to the EasyScience project <https://github.com/easyScience/EasyScience
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+
+from ..io import SerializerBase
+from ..variable import DescriptorNumber
+from ..variable import Parameter
+from .new_base import NewBase
+
+
+class ModelBase(NewBase):
+    """
+    This is the base class for all model classes in EasyScience.
+    It provides methods to get parameters for fitting and analysis as well as proper serialization/deserialization for
+    DescriptorNumber/Parameter attributes.
+
+    It assumes that Parameters/DescriptorNumbers are assigned as properties with the getters returning the parameter
+    but the setter only setting the value of the parameter.
+    e.g.
+    ```python
+    @property
+    def my_param(self) -> Parameter:
+        return self._my_param
+
+    @my_param.setter
+    def my_param(self, new_value: float) -> None:
+        self._my_param.value = new_value
+    ```
+    """
+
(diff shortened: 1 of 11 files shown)


src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+        if skip is None:
+            skip = []
+        if self._default_unique_name and 'unique_name' not in skip:
+            skip.append('unique_name')
+        if self._display_name is None:
+            skip.append('display_name')
+        return serializer._convert_to_dict(self, skip=skip, full_encode=False)
+
+    @classmethod
+    def from_dict(cls, obj_dict: Dict[str, Any]) -> None:
```
rozyczko: Incorrect return type. Should be `NewBase` or `self`

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
+from __future__ import annotations
```
rozyczko, line 1: Not needed in >3.10

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+
+from ..global_object.undo_redo import property_stack
+from ..io.serializer_base import SerializerBase
+
+
+class NewBase:
+    """
+    This is the new base class for easyscience objects.
+    It provides serialization capabilities as well as unique naming and display naming.
+    """
```
rozyczko, line 27: You might consider using `__slots__` for memory efficiency. This tells the interpreter which attributes the instances are allowed to have. This also removes `__dict__`, which is normally used to store attributes dynamically.
Here, it would be 
```python
__slots__ =  ['_global_object', '_unique_name', '_default_unique_name', '_display_name'] 
```

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+        self._display_name = name
+
+    def to_dict(self, skip: Optional[List[str]] = None) -> Dict[str, Any]:
+        """
+        Convert an EasyScience object into a full dictionary using `SerializerBase`s generic `convert_to_dict` method.
+
+        :param skip: List of field names as strings to skip when forming the dictionary
+        :return: encoded object containing all information to reform an EasyScience object.
+        """
+        serializer = SerializerBase()
```
rozyczko, line 101: can we instantiate `SerializerBase` just once in the init?

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+        """Return a copy of the object."""
+        temp = self.to_dict(skip=['unique_name'])
+        new_obj = self.__class__.from_dict(temp)
+        return new_obj
+
+    def __deepcopy__(self, memo):
+        return self.__copy__()
+
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__} `{self.unique_name}`'
```
rozyczko, line 145: Will we not need `__reduce__` for the base class? This might come in handy at some point.
(Yes, this is asking for overengineering of options we don't strictly need...)

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+
+class NewBase:
+    """
+    This is the new base class for easyscience objects.
+    It provides serialization capabilities as well as unique naming and display naming.
+    """
+
+    def __init__(self, unique_name: Optional[str] = None, display_name: Optional[str] = None):
+        self._global_object = global_object
+        if unique_name is None:
```
rozyczko, line 31: Should we not validate `unique_name` here? We have access to the global map so we can check if `unique_name` is indeed unique

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def get_fit_parameters(self) -> List[Parameter]:
+        """
+        This is an alias for `get_free_parameters`.
+        To be removed when fully moved to new base classes and minimizer can be changed.
+        """
+        return self.get_free_parameters()
+
+    @classmethod
+    def from_dict(cls, obj_dict: Dict[str, Any]) -> None:
```
rozyczko: Again, wrong return type

src/easyscience/base_classes/model_base.py
```diff
@@ -0,0 +1,115 @@
...
+            kwargs = SerializerBase.deserialize_dict(obj_dict)
+            parameter_placeholder = {}
+            for key, value in kwargs.items():
+                if isinstance(value, DescriptorNumber):
+                    parameter_placeholder[key] = value
+                    kwargs[key] = value.value
+            cls_instance = cls(**kwargs)
+            for key, value in parameter_placeholder.items():
+                temp_param = getattr(cls_instance, key)
+                setattr(cls_instance, '_' + key, value)
```
rozyczko: `getattr` and `setattr` could fail. Consider wrapping them in try..except

rozyczko (commented):
minor comments, but the whole structure looks sound

src/easyscience/base_classes/new_base.py
damskii9992: This is only deprecated in 3.14:
https://peps.python.org/pep-0749/#abstract
For now it is still used, since Ruff is more aggressive in its suggestions with it enabled:
https://github.com/astral-sh/ruff/issues/18502
When 3.14 becomes our minimally supported version, then we can get rid of it, as dicussed here:
https://github.com/astral-sh/ruff/issues/20739

src/easyscience/base_classes/new_base.py
damskii9992: That should probably be an entire ADR, as using `__slots__` will disallow us from dynamically making new class attributes since `__dict__` is removed, which essentially forces us to use `__slots__` everywhere . . .
For example, we would have to make slots for `__serializer_id`, `_dependency_map_serializer_ids` etc. I.e. any attribute we create, even transient ones, would need to have an entry in `__slots__`.
(It was pretty interesting reading up on this, so thanks for that I guess).
Another issue with this is for multiple inheritance/mixins, where both classes can't have `__slots__`.

damskii9992: We **can** include `__dict__` in `__slots__` which would retain the ability for dynamic attribute creation, but we'd lose out on some of the memory efficiency and the faster attribute access for those attributes.

damskii9992: Made the ADR here: #165. For discussion and for future reference. With the findings I found when researching `__slots__`

src/easyscience/base_classes/new_base.py
damskii9992: We can.
But that would slow down class instantiation. I actually prefer to have it here, as `to_dict` should only be called once, if even at all. So at best we get a speed-up by only instantiating it when we need it, and at worst we get the same total time. Unless we of course serialize multiple times.
I also thought about having it in the `global_object`, as we only really need 1 instance of the class globally.
I also thought about making it simple functions, since most of the methods are static functions, which hints that it shouldn't even be a class.

src/easyscience/base_classes/new_base.py
damskii9992: I thought about it, since it exists in `BasedBase`. But I ended up at the conclusion that we shouldn't make methods we don't need. We can always add it later if we find out we want to make our classes pickable.

src/easyscience/base_classes/new_base.py
damskii9992: The `unique_name` gets validated in:
```python
self._global_object.map.add_vertex(self, obj_type='created')
```
I.e. when adding it to the global map.

src/easyscience/base_classes/new_base.py
rozyczko: Keeping `SerializerBase` makes sense - all serialization related methods are in one location and there are enough instance methods to justify it being a separate class. If we take all the static methods away and plop them into say, `serialization_utils`, we would have to import them separately later and remember about that...

src/easyscience/base_classes/model_base.py
damskii9992: They should only fail if we have implemented our classes incorrectly, in which case we should let them fail and throw an error to know about it, or if the serialized dict has been tampered with, in which case other pieces could fail anyways and we should also just let it fail.
Do you have a good reason as to why this could fail otherwise?

src/easyscience/base_classes/new_base.py
damskii9992: It's only `_convert_to_dict` and its utility method `_recursive_encoder` which are not static methods. And both of those could just as easily have been made static methods. The class doesn't even have an `__init__` or any attributes . . .

src/easyscience/base_classes/model_base.py
rozyczko: e.g. to let users know that the failure stems from this issue. Bubbled exceptions often lose context and it's difficult to pinpoint what happened, if you're in a library code

src/easyscience/base_classes/new_base.py
rozyczko: I don't think `_convert_to_dict` can be easily made static... 

`_recursive_encoder` - yes. but not the other one.

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+            raise TypeError('Display name must be a string or None')
+        self._display_name = display_name
+
+    @property
+    def _arg_spec(self) -> Set[str]:
+        """
+        This method is used by the serializer to determine which arguments are needed
+        by the constructor to deserialize the object.
+        """
+        spec = getfullargspec(self.__class__.__init__)
```
rozyczko: It seems `inspect.signature()` is the modern/updated version of `getfullargspec`. The latter was supposed to be deprecated in >3.7 but was left for backward compatibility. We should move to modern/maintained calls.

damskii9992: I replaced `getfullargspec` where it was somewhat simple to do an opened an issue on replacing it in the `SerializerBases` `_convert_to_dict` method here: #166

src/easyscience/base_classes/model_base.py
damskii9992: Is this (the new code) what you had in mind?

src/easyscience/base_classes/new_base.py
seventil: _convert_to_dict only uses self to later refer where to pull `_convert_to_dict` and `_recursive_encoder` from. If we only use these functions from one SerializerBase encoder - they can be static. And not just static but even plain funcitons.

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+        :return: The pretty display name.
+        """
+        display_name = self._display_name
+        if display_name is None:
+            display_name = self.unique_name
+        return display_name
+
+    @display_name.setter
+    @property_stack
+    def display_name(self, name: str) -> None:
```
seventil: it think name hint should be optional

src/easyscience/base_classes/new_base.py
```diff
@@ -0,0 +1,145 @@
...
+        new_class_objs = list(k for k in dir(self.__class__) if not k.startswith('_'))
+        return sorted(new_class_objs)
+
+    def __copy__(self) -> NewBase:
+        """Return a copy of the object."""
+        temp = self.to_dict(skip=['unique_name'])
+        new_obj = self.__class__.from_dict(temp)
+        return new_obj
+
+    def __deepcopy__(self, memo):
```
seventil, line 141: my linter complains for `memo` not used with the quick fix to del memo in the body of the method

rozyczko: https://stackoverflow.com/questions/69469980/what-is-an-opaque-object-in-the-context-of-the-copy-module

src/easyscience/base_classes/new_base.py
damskii9992: ??? WDYM?

src/easyscience/base_classes/new_base.py
seventil: But in this context, memo is not used. I understand that it's necessary to have as a parameter, but the fact that it's not used in the method body leads to a linter problem. `del memo` is a common easy fix to dealing with unused function parameters.

tbh i don't think that's an issue, just a reminder in case `ruff check` fails here or smth

PR #172, opened 2025-12-03
New calculator factory

rozyczko:
This PR implements the architecture decision from [Discussion #160](vscode-file://vscode-app/d:/util/VSCodeI/9ba40f8204/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to introduce a new stateless calculator factory pattern for physics calculators in EasyScience.

Key Components Added

Component	|File	| Purpose
--- | --- | ---
CalculatorBase | calculator_base.py | Abstract base class for physics calculators that compute theoretical results
CalculatorFactoryBase | calculator_factory.py | Abstract factory for creating calculator instances
SimpleCalculatorFactory | calculator_factory.py | Concrete factory implementation using dictionary registry
ModelCollection | model_collection.py | Generic collection class for NewBase/ModelBase objects

Files Modified

File | Change
--- | --- 
map.py:70-76 | Made _store private (__store) to enforce encapsulation
parameter.py:1025-1029 | Updated to use public API instead of private _store access
collection_base.py:64-68 | Extended type checking to include NewBase
__init__.py | Exported new classes with backwards compatibility note for InterfaceFactoryTemplate
__init__.py | Added ModelCollection export


src/easyscience/base_classes/model_collection.py
```diff
@@ -0,0 +1,286 @@
...
+class ModelCollection(NewBase, MutableSequence[T]):
+    """
+    A collection class for NewBase/ModelBase objects.
+    This provides list-like functionality while maintaining EasyScience features
+    like serialization and interface bindings.
+    """
+
+    def __init__(
+        self,
+        name: str,
```
henrikjacobsenfys: Do we still want both `name` and `display_name`? I thought one of the goals of NewBase was to get rid of `name`?

src/easyscience/base_classes/model_collection.py
```diff
@@ -0,0 +1,286 @@
...
+    def _remove_item(self, item: NewBase) -> None:
+        """Remove an item from the collection and clean up graph edges."""
+        self._global_object.map.prune_vertex_from_edge(self, item)
+
+    @property
+    def name(self) -> str:
+        """Get the name of the collection."""
+        return self._name
+
+    @name.setter
```
henrikjacobsenfys: I suppose if we're keeping name it should verify that the input is a string?

src/easyscience/base_classes/model_collection.py
```diff
@@ -0,0 +1,286 @@
...
+        self._name = new_name
+
+    @property
+    def interface(self) -> Optional[InterfaceFactoryTemplate]:
+        """Get the current interface of the collection."""
+        return self._interface
+
+    @interface.setter
+    def interface(self, new_interface: Optional[InterfaceFactoryTemplate]) -> None:
+        """Set the interface and propagate to all items."""
```
henrikjacobsenfys: And some verification here?

src/easyscience/base_classes/model_collection.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # Use @overload to provide precise type hints for different __getitem__ argument types
+    @overload
+    def __getitem__(self, idx: int) -> T: ...
+    @overload
+    def __getitem__(self, idx: slice) -> 'ModelCollection[T]': ...
+    @overload
+    def __getitem__(self, idx: str) -> T: ...
+
+    def __getitem__(self, idx: Union[int, slice, str]) -> Union[T, 'ModelCollection[T]']:
```
henrikjacobsenfys: Should we gradually move towards the "new" way of type hinting with `|` instead of `Union`?

src/easyscience/base_classes/model_collection.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def get_all_variables(self) -> List[Any]:
+        """Get all variables from all items in the collection."""
+        variables: List[Any] = []
+        for item in self._data:
+            if hasattr(item, 'get_all_variables'):
+                variables.extend(item.get_all_variables())  # type: ignore[attr-defined]
+        return variables
+
+    def get_all_parameters(self) -> List[Any]:
```
henrikjacobsenfys: I don't think this is needed, since the `NewBase` `get_all_parameters` method uses `get_all_variables` and selects the correct ones from there. Same for `get_fit_parameters`

henrikjacobsenfys (commented):
Just a few questions, otherwise looks good to me

src/easyscience/base_classes/model_collection.py
damskii9992: Yes, we should . . .

src/easyscience/fitting/calculators/calculator_base.py
```diff
@@ -0,0 +1,266 @@
...
+            def calculate(self, x):
+                # Use self._model and self._instrumental_parameters
+                return computed_values
+    """
+
+    name: str = 'base'
+
+    def __init__(
+        self,
+        model: NewBase,
```
damskii9992, line 87: The typehint should be ModelBase for now and should probably be `SampleModelBase` when that base class has been properly updated.

src/easyscience/fitting/calculators/calculator_base.py
```diff
@@ -0,0 +1,266 @@
...
+            Instrumental parameters that affect the calculation, such as
+            resolution, wavelength, or detector settings.
+        unique_name : str, optional
+            Unique identifier for this calculator instance.
+        display_name : str, optional
+            Human-readable name for display purposes.
+        **kwargs : Any
+            Additional calculator-specific options.
+        """
+        if model is None:
```
damskii9992, line 111: why not `if not isinstance(ModelBase)`?

src/easyscience/fitting/calculators/calculator_base.py
```diff
@@ -0,0 +1,266 @@
...
+
+        # Initialize NewBase with naming
+        super().__init__(unique_name=unique_name, display_name=display_name)
+
+        self._model = model
+        self._instrumental_parameters = instrumental_parameters
+        self._additional_kwargs = kwargs
+        # Register this calculator and model in the global object map
+        if hasattr(model, 'unique_name'):
+            self._global_object.map.add_edge(self, model)
```
damskii9992, line 122: So, if we know that the model is a `ModelBase` (or even `NewBase`), then we know it has a `unique_name` attribute and this check is redundant. But also, why add the edge? As far as I know, we don't use these "edges" in the map anywhere, and I hope to remove it entirely, so unless you have a reason to add it, please don't.

src/easyscience/fitting/calculators/calculator_factory.py
```diff
@@ -0,0 +1,324 @@
...
+    def available_calculators(self) -> List[str]:
+        """
+        Return a list of available calculator names.
+
+        Returns
+        -------
+        List[str]
+            Names of all calculators that can be created by this factory.
+        """
+        ...
```
damskii9992, line 103: I think we can do better than this in the base class. How about if the factory has an internal attribute `_available_calculators` which is an empty list and a generic method which attempts to import a package based on input, and if successful adds it to this list. That way we can have automatic granularity of installed calculators and even if a calculator for some reason fails on the users end we can still support the other calculators.

src/easyscience/variable/parameter.py
```diff
@@ -1029,7 +1029,8 @@ def resolve_pending_dependencies(self) -> None:
 
     def _find_parameter_by_serializer_id(self, serializer_id: str) -> Optional['DescriptorNumber']:
         """Find a parameter by its serializer_id from all parameters in the global map."""
-        for obj in self._global_object.map._store.values():
+        for key in self._global_object.map.vertices():
+            obj = self._global_object.map.get_item_by_key(key)
```
damskii9992, line 1033: Haha, I knew I remembered seeing that you used the internal `_store` somewhere :D

damskii9992 (changes requested):
Can you please split this PR into two? I believe the ModelCollection needs its own PR and ADR discussion.

rozyczko:
> Can you please split this PR into two? I believe the ModelCollection needs its own PR and ADR discussion.

I can but the subsequent merge of these two would be ugly as the functionality is overlapping to some extent.

src/easyscience/variable/parameter.py
rozyczko: shhh... you didn't SEE ANYTHING

rozyczko:
split and closed with references

PR #174, opened 2025-12-08
Updated composite actions

rozyczko:
Moved the pixi install action to a central repo in the `actions` project.

Should be helpful for projects inheriting this and other, related actions.


damskii9992 (approved):
LGTM

PR #175, opened 2025-12-13
added thread lock to avoid conflicts between GC and parameter

### LINKED ISSUE #100: RuntimeError: dictionary changed size during iteration
```
=================================== FAILURES ===================================
____________________ test_Lattice_from_pars[hexagonal-rad] _____________________
[gw0] darwin -- Python 3.13.1 /Library/Frameworks/Python.framework/Versions/3.13/bin/python

value = [3, 3, 4, np.float64(1.5707963267948966), np.float64(1.5707963267948966), np.float64(2.0943951023931953)]
ang_unit = 'rad'

    @pytest.mark.parametrize("ang_unit", ("deg", "rad"))
    @pytest.mark.parametrize("value", basic_pars)
    def test_Lattice_from_pars(value: list, ang_unit: str):
        ref = [v for v in value]
    
        if ang_unit == "rad":
            value = [
                value[0],
                value[1],
                value[2],
                np.deg2rad(value[3]),
                np.deg2rad(value[4]),
                np.deg2rad(value[5]),
            ]
    
>       l = Lattice(*value, ang_unit=ang_unit)

tests/unit_tests/components/test_Lattice.py:123: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easycrystallography/Components/Lattice.py:106: in __init__
    self.angle_beta = angle_beta
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/Objects/ObjectClasses.py:327: in __setattr__
    super(BaseObj, self).__setattr__(key, value)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/global_object/hugger/property.py:92: in __set__
    return super().__set__(instance, value)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/Objects/ObjectClasses.py:350: in setter
    obj._kwargs[key].value = value
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/global_object/undo_redo.py:473: in wrapper
    global_object.stack.push(PropertyStack(obj, func, old_value, new_value, **kwargs))
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/global_object/undo_redo.py:191: in push
    command.redo()
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/global_object/undo_redo.py:337: in redo
    self._set_func(self._parent, self._new_value)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/Objects/variable/parameter.py:191: in value
    value = self._constraint_runner(self.builtin_constraints, self._scalar.value)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/Objects/variable/parameter.py:394: in _constraint_runner
    constained_value = constraint(no_set=True)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/Constraints.py:115: in __call__
    dependent_obj = self.get_obj(self.dependent_obj_ids)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/Constraints.py:157: in get_obj
    return self._global_object.map.get_item_by_key(key)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easyscience/global_object/map.py:109: in get_item_by_key
    if item_id in self._store.keys():
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <WeakValueDictionary at 0x105468550>

    def keys(self):
        if self._pending_removals:
            self._commit_removals()
        with _IterationGuard(self):
>           for k, wr in self.data.items():
E           RuntimeError: dictionary changed size during iteration

/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/weakref.py:222: RuntimeError
=============================== warnings summary ===============================
tests/unit_tests/components/test_Lattice.py: 24 warnings
  /Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easycrystallography/Components/Lattice.py:133: DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments. To learn more, see the migration guide https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
    matrix = np.array(matrix, dtype=np.float64).reshape((3, 3))

tests/unit_tests/components/test_Lattice.py: 24 warnings
  /Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/easycrystallography/Components/Lattice.py:133: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
    matrix = np.array(matrix, dtype=np.float64).reshape((3, 3))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/unit_tests/components/test_Lattice.py::test_Lattice_from_pars[hexagonal-rad] - RuntimeError: dictionary changed size during iteration
```

rozyczko:
Fixed the `RuntimeError: dictionary changed size during iteration` issue in the vertices method by implementing thread-safety: iterations handling modifications during garbage collection.
This seems to have been a concurrency problem with weak references, where the dictionary is being modified during iteration.

Take the attached script (.txt -> .py cause GitHub)
[run_test_suite.txt](https://github.com/user-attachments/files/22681211/run_test_suite.txt)

Run it in say, `EasyReflectometryLib` or `corelib`

This now runs fine.

### Root Cause

The `Map` class uses `weakref.WeakValueDictionary()` to store object references. When:

1. `generate_unique_name()` is called (e.g., when creating a new `Parameter`)
2. It calls `self.map.vertices()` which returns `list(self._store.keys())`
3. During the iteration over `.keys()`, the garbage collector may run
4. GC cleans up objects with no remaining strong references
5. This removes entries from the `WeakValueDictionary`
6. Python raises `RuntimeError: dictionary changed size during iteration`

This is particularly likely to happen when:
- Many temporary Parameter objects are created (e.g., in `Lattice.volume` property)
- The GC runs during the iteration
- Multiple objects are being created/destroyed in quick succession

### DIFF
--- a/src/easyscience/global_object/map.py
+++ b/src/easyscience/global_object/map.py
@@ -4,6 +4,7 @@
 
 import gc
 import sys
+import threading
 import weakref
 from typing import List
 from typing import Optional
@@ -73,10 +74,30 @@ class Map:
         self._store = weakref.WeakValueDictionary()
         # A dict with object names as keys and a list of their object types as values, with weak references
         self.__type_dict = {}
+        # Lock for thread-safe access to _store
+        self._lock = threading.RLock()
 
     def vertices(self) -> List[str]:
-        """returns the vertices of a map"""
-        return list(self._store.keys())
+        """Returns the vertices of a map.
+
+        Uses a threading lock to prevent RuntimeError from WeakValueDictionary
+        being modified during iteration.
+        """
+        with self._lock:
+            return list(self._store.keys())
+
+        # --- Option 1: Retry loop ---
+        # """Returns the vertices of a map.
+        #
+        # Uses a retry loop to handle RuntimeError that can occur when the
+        # WeakValueDictionary is modified during iteration (e.g., by garbage collection).
+        # """
+        # while True:
+        #     try:
+        #         return list(self._store.keys())
+        #     except RuntimeError:
+        #         # Dictionary changed size during iteration, retry
+        #         continue
 
     def edges(self):
         """returns the edges of a map"""
@@ -103,13 +124,42 @@ class Map:
         return [key for key, item in self.__type_dict.items() if obj_type in item.type]
 
     def get_item_by_key(self, item_id: str) -> object:
-        if item_id in self._store.keys():
-            return self._store[item_id]
-        raise ValueError('Item not in map.')
+        with self._lock:
+            if item_id in self._store.keys():
+                return self._store[item_id]
+            raise ValueError('Item not in map.')
+
+        # --- Option 1: Retry loop ---
+        # while True:
+        #     try:
+        #         if item_id in self._store.keys():
+        #             return self._store[item_id]
+        #         raise ValueError('Item not in map.')
+        #     except RuntimeError:
+        #         # Dictionary changed size during iteration, retry
+        #         continue
 
     def is_known(self, vertex: object) -> bool:
-        # All objects should have a 'unique_name' attribute
-        return vertex.unique_name in self._store.keys()
+        """Check if a vertex is known in the map.
+
+        All objects should have a 'unique_name' attribute.
+        Uses a threading lock to prevent RuntimeError from WeakValueDictionary.
+        """
+        with self._lock:
+            return vertex.unique_name in self._store.keys()
+
+        # --- Option 1: Retry loop ---
+        # """Check if a vertex is known in the map.
+        #
+        # All objects should have a 'unique_name' attribute.
+        # Uses a retry loop to handle RuntimeError from WeakValueDictionary.
+        # """
+        # while True:
+        #     try:
+        #         return vertex.unique_name in self._store.keys()
+        #     except RuntimeError:
+        #         # Dictionary changed size during iteration, retry
+        #         continue
 
     def find_type(self, vertex: object) -> List[str]:
         if self.is_known(vertex):
@@ -124,13 +174,14 @@ class Map:
             self.__type_dict[obj.unique_name].type = new_type
 
     def add_vertex(self, obj: object, obj_type: str = None):
-        name = obj.unique_name
-        if name in self._store.keys():
-            raise ValueError(f'Object name {name} already exists in the graph.')
-        self._store[name] = obj
-        self.__type_dict[name] = _EntryList()  # Add objects type to the list of types
-        self.__type_dict[name].finalizer = weakref.finalize(self._store[name], self.prune, name)
-        self.__type_dict[name].type = obj_type
+        with self._lock:
+            name = obj.unique_name
+            if name in self._store.keys():
+                raise ValueError(f'Object name {name} already exists in the graph.')
+            self._store[name] = obj
+            self.__type_dict[name] = _EntryList()  # Add objects type to the list of types
+            self.__type_dict[name].finalizer = weakref.finalize(self._store[name], self.prune, name)
+            self.__type_dict[name].type = obj_type
 
     def add_edge(self, start_obj: object, end_obj: object):
         if start_obj.unique_name in self.__type_dict.keys():
@@ -167,9 +218,11 @@ class Map:
             del self.__type_dict[vertex1][self.__type_dict[vertex1].index(vertex2)]
 
     def prune(self, key: str):
-        if key in self.__type_dict.keys():
-            del self.__type_dict[key]
-            del self._store[key]
+        with self._lock:
+            if key in self.__type_dict.keys():
+                del self.__type_dict[key]
+                if key in self._store:
+                    del self._store[key]
 
     def find_isolated_vertices(self) -> list:
         """returns a list of isolated vertices."""
@@ -262,10 +315,11 @@ class Map:
 
     def _clear(self):
         """Reset the map to an empty state. Only to be used for testing"""
-        for vertex in self.vertices():
-            self.prune(vertex)
-        gc.collect()
-        self.__type_dict = {}
+        with self._lock:
+            for vertex in self.vertices():
+                self.prune(vertex)
+            gc.collect()
+            self.__type_dict = {}
 
     def __repr__(self) -> str:
         return f'Map object of {len(self._store)} vertices.'
--- a/src/easyscience/global_object/map.py
+++ b/src/easyscience/global_object/map.py
@@ -4,7 +4,6 @@
 
 import gc
 import sys
-import threading
 import weakref
 from typing import List
 from typing import Optional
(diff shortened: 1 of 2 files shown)


rozyczko:
Nooope. Going back to the thread lock.

gc.disable() affects the *GLOBAL* garbage collector state across the whole process (all threads).
We use `corelib` in libraries (same thread) and in GUIs (in a separate - non-GUI thread).
Disabling GC in a GUI thread can have unintented consequences, the least being slowing down due to memory buildup.

damskii9992:
> gc.disable() affects the _GLOBAL_ garbage collector state across the whole process (all threads). We use `corelib` in libraries (same thread) and in GUIs (in a separate - non-GUI thread). Disabling GC in a GUI thread can have unintented consequences, the least being slowing down due to memory buildup.

Did you encounter issues in the application when you tried it out? It is only a (VERY) temporary disabling of GC, so memory buildup shouldn't occur.

src/easyscience/global_object/map.py
```diff
@@ -103,13 +124,42 @@ def _nested_get(self, obj_type: str) -> List[str]:
         return [key for key, item in self.__type_dict.items() if obj_type in item.type]
 
     def get_item_by_key(self, item_id: str) -> object:
-        if item_id in self._store.keys():
-            return self._store[item_id]
-        raise ValueError('Item not in map.')
+        with self._lock:
+            if item_id in self._store.keys():
+                return self._store[item_id]
+            raise ValueError('Item not in map.')
```
damskii9992: We should be able to avoid this entire error if we just use normal introspection of the dictionary rather than first get the keys:
```python
if item_id in self_store:
   return self._store[item_id]
```
Then we're not iterating over values and thus don't need this (it should also be substantially faster).

rozyczko:
> > gc.disable() affects the _GLOBAL_ garbage collector state across the whole process (all threads). We use `corelib` in libraries (same thread) and in GUIs (in a separate - non-GUI thread). Disabling GC in a GUI thread can have unintented consequences, the least being slowing down due to memory buildup.
> 
> Did you encounter issues in the application when you tried it out? It is only a (VERY) temporary disabling of GC, so memory buildup shouldn't occur.

Doesn't matter. This is still a valid issue - we shouldn't do things in the library, which can globally affect behaviour of the code linked to it?

src/easyscience/global_object/map.py
```diff
@@ -103,13 +124,42 @@ def _nested_get(self, obj_type: str) -> List[str]:
...
     def is_known(self, vertex: object) -> bool:
-        # All objects should have a 'unique_name' attribute
-        return vertex.unique_name in self._store.keys()
+        """Check if a vertex is known in the map.
+
+        All objects should have a 'unique_name' attribute.
+        Uses a threading lock to prevent RuntimeError from WeakValueDictionary.
+        """
+        with self._lock:
+            return vertex.unique_name in self._store.keys()
```
damskii9992: Again, we can avoid the lock and this issue if we don't call the keys() method. Just use regular inspection.

src/easyscience/global_object/map.py
```diff
@@ -124,13 +174,14 @@ def change_type(self, obj, new_type: str):
...
-        name = obj.unique_name
-        if name in self._store.keys():
-            raise ValueError(f'Object name {name} already exists in the graph.')
-        self._store[name] = obj
-        self.__type_dict[name] = _EntryList()  # Add objects type to the list of types
-        self.__type_dict[name].finalizer = weakref.finalize(self._store[name], self.prune, name)
-        self.__type_dict[name].type = obj_type
+        with self._lock:
+            name = obj.unique_name
+            if name in self._store.keys():
```
damskii9992: Also here

src/easyscience/global_object/map.py
```diff
@@ -167,9 +218,11 @@ def prune_vertex_from_edge(self, parent_obj, child_obj):
             del self.__type_dict[vertex1][self.__type_dict[vertex1].index(vertex2)]
 
     def prune(self, key: str):
-        if key in self.__type_dict.keys():
```
damskii9992, line 170: And again here.

src/easyscience/global_object/map.py
```diff
@@ -262,10 +315,11 @@ def is_connected(self, vertices_encountered=None, start_vertex=None) -> bool:
...
         """Reset the map to an empty state. Only to be used for testing"""
-        for vertex in self.vertices():
-            self.prune(vertex)
-        gc.collect()
-        self.__type_dict = {}
+        with self._lock:
+            for vertex in self.vertices():
+                self.prune(vertex)
+            gc.collect()
+            self.__type_dict = {}
```
damskii9992: Actually, can't we get rid of this method and just use the built-in `clear` method of a regular dictionary?

src/easyscience/global_object/map.py
rozyczko: '''
• Key existence vs. value availability:
item_id in self._store checks only if the key is present. If the value has been collected, the key will also be removed, so you won’t get a dangling key.
• Iteration safety:
Iterating over self._store.keys() or items() is safe, but note that the contents may shrink during iteration if garbage collection occurs.
• Direct access:
self._store[item_id] will raise KeyError if the object was collected.
'''

So even with direct query we will get the same issue.

rozyczko: Actually you might be right.
Querying the dict directly with the KeyError except check _should_ be enough. Let's check this with the test script

damskii9992: Wait, was that a ChatGPT prompt output? It obviously hallucinates. I literally said that its safe to iterate over self._store.keys(), which is what is causing our error.

rozyczko: Yes, this is what you get when you take AI verbatim. Some good points but all covered in nonsense.

damskii9992:
I really like the retry look implementation for the vertices method. If this works I'd say its' good to merge :) 
Although it looks like codecov wants more tests.

PR #176, opened 2025-12-16
Mkdocs conversion

rozyczko:
Draft PR to expose the issue rather than directly merge.
@AndrewSazonov 's modifications in another branch (`new_docs`) should be either merged here or the other way around.

This branch is what's currently published on https://easyscience.github.io/corelib/


rozyczko:
This isn't getting merged but the branch is staying for updates on Andrew's work.

PR #188, opened 2026-01-09
Added desired units and tests

### LINKED ISSUE #187: Consider adding a desired unit to dependent parameters
Resolving dependency expressions can lead to units being converted to SI units. Users will often want non-SI units such as `meV` and `angstrom`

henrikjacobsenfys:
Added desired units for the dependent parameter constructors, and a method to change it later: `set_desired_unit`

### DIFF
--- a/src/easyscience/variable/parameter.py
+++ b/src/easyscience/variable/parameter.py
@@ -36,13 +36,13 @@ class Parameter(DescriptorNumber):
     # Used by serializer
     # We copy the parent's _REDIRECT and modify it to avoid altering the parent's class dict
     _REDIRECT = DescriptorNumber._REDIRECT.copy()
-    _REDIRECT['callback'] = None
+    _REDIRECT["callback"] = None
 
     def __init__(
         self,
         name: str,
         value: numbers.Number,
-        unit: Optional[Union[str, sc.Unit]] = '',
+        unit: Optional[Union[str, sc.Unit]] = "",
         variance: Optional[numbers.Number] = 0.0,
         min: Optional[numbers.Number] = -np.inf,
         max: Optional[numbers.Number] = np.inf,
@@ -76,26 +76,30 @@ class Parameter(DescriptorNumber):
             Undo/Redo functionality is implemented for the attributes `value`, `variance`, `error`, `min`, `max`, `bounds`, `fixed`, `unit`
         """  # noqa: E501
         # Extract and ignore serialization-specific fields from kwargs
-        kwargs.pop('_dependency_string', None)
-        kwargs.pop('_dependency_map_serializer_ids', None)
-        kwargs.pop('_independent', None)
+        kwargs.pop("_dependency_string", None)
+        kwargs.pop("_dependency_map_serializer_ids", None)
+        kwargs.pop("_independent", None)
 
         if not isinstance(min, numbers.Number):
-            raise TypeError('`min` must be a number')
+            raise TypeError("`min` must be a number")
         if not isinstance(max, numbers.Number):
-            raise TypeError('`max` must be a number')
+            raise TypeError("`max` must be a number")
         if not isinstance(value, numbers.Number):
-            raise TypeError('`value` must be a number')
+            raise TypeError("`value` must be a number")
         if value < min:
-            raise ValueError(f'{value=} can not be less than {min=}')
+            raise ValueError(f"{value=} can not be less than {min=}")
         if value > max:
-            raise ValueError(f'{value=} can not be greater than {max=}')
+            raise ValueError(f"{value=} can not be greater than {max=}")
         if np.isclose(min, max, rtol=1e-9, atol=0.0):
-            raise ValueError('The min and max bounds cannot be identical. Please use fixed=True instead to fix the value.')
+            raise ValueError(
+                "The min and max bounds cannot be identical. Please use fixed=True instead to fix the value."
+            )
         if not isinstance(fixed, bool):
-            raise TypeError('`fixed` must be either True or False')
+            raise TypeError("`fixed` must be either True or False")
         self._independent = True
-        self._fixed = fixed  # For fitting, but must be initialized before super().__init__
+        self._fixed = (
+            fixed  # For fitting, but must be initialized before super().__init__
+        )
         self._min = sc.scalar(float(min), unit=unit)
         self._max = sc.scalar(float(max), unit=unit)
 
@@ -139,7 +143,7 @@ class Parameter(DescriptorNumber):
         :return: A new dependent Parameter object.
         """  # noqa: E501
         # Set default values for required parameters for the constructor, they get overwritten by the dependency anyways
-        default_kwargs = {'value': 0.0, 'variance': 0.0, 'min': -np.inf, 'max': np.inf}
+        default_kwargs = {"value": 0.0, "variance": 0.0, "min": -np.inf, "max": np.inf}
         # Update with user-provided kwargs, to avoid errors.
         default_kwargs.update(kwargs)
         parameter = cls(name=name, **default_kwargs)
@@ -156,15 +160,21 @@ class Parameter(DescriptorNumber):
         """
         if not self._independent:
             # Update the value of the parameter using the dependency interpreter
-            temporary_parameter = self._dependency_interpreter(self._clean_dependency_string)
+            temporary_parameter = self._dependency_interpreter(
+                self._clean_dependency_string
+            )
             self._scalar.value = temporary_parameter.value
             self._scalar.unit = temporary_parameter.unit
             self._scalar.variance = temporary_parameter.variance
             self._min.value = (
-                temporary_parameter.min if isinstance(temporary_parameter, Parameter) else temporary_parameter.value
+                temporary_parameter.min
+                if isinstance(temporary_parameter, Parameter)
+                else temporary_parameter.value
             )  # noqa: E501
             self._max.value = (
-                temporary_parameter.max if isinstance(temporary_parameter, Parameter) else temporary_parameter.value
+                temporary_parameter.max
+                if isinstance(temporary_parameter, Parameter)
+                else temporary_parameter.value
             )  # noqa: E501
             self._min.unit = temporary_parameter.unit
             self._max.unit = temporary_parameter.unit
@@ -174,7 +184,7 @@ class Parameter(DescriptorNumber):
 
             self._notify_observers()
         else:
-            warnings.warn('This parameter is not dependent. It cannot be updated.')
+            warnings.warn("This parameter is not dependent. It cannot be updated.")
 
     def make_dependent_on(
         self,
@@ -207,22 +217,24 @@ class Parameter(DescriptorNumber):
 
         """  # noqa: E501
         if not isinstance(dependency_expression, str):
-            raise TypeError('`dependency_expression` must be a string representing a valid dependency expression.')
+            raise TypeError(
+                "`dependency_expression` must be a string representing a valid dependency expression."
+            )
         if not (isinstance(dependency_map, dict) or dependency_map is None):
             raise TypeError(
-                '`dependency_map` must be a dictionary of dependencies and their'
-                'corresponding names in the dependecy expression.'
+                "`dependency_map` must be a dictionary of dependencies and their"
+                "corresponding names in the dependecy expression."
             )  # noqa: E501
         if isinstance(dependency_map, dict):
             for key, value in dependency_map.items():
                 if not isinstance(key, str):
                     raise TypeError(
-                        '`dependency_map` keys must be strings representing the names of'
-                        'the dependencies in the dependency expression.'
+                        "`dependency_map` keys must be strings representing the names of"
+                        "the dependencies in the dependency expression."
                     )  # noqa: E501
                 if not isinstance(value, DescriptorNumber):
                     raise TypeError(
-                        f'`dependency_map` values must be DescriptorNumbers or Parameters. Got {type(value)} for {key}.'
+                        f"`dependency_map` values must be DescriptorNumbers or Parameters. Got {type(value)} for {key}."
                     )  # noqa: E501
 
         # If we're overwriting the dependency, store the old attributes
@@ -230,11 +242,11 @@ class Parameter(DescriptorNumber):
         self._previous_independent = self._independent
         if not self._independent:
             self._previous_dependency = {
-                '_dependency_string': self._dependency_string,
-                '_dependency_map': self._dependency_map,
-                '_dependency_interpreter': self._dependency_interpreter,
-                '_clean_dependency_string': self._clean_dependency_string,
-                '_desired_unit': self._desired_unit,
+                "_dependency_string": self._dependency_string,
+                "_dependency_map": self._dependency_map,
+                "_dependency_interpreter": self._dependency_interpreter,
+                "_clean_dependency_string": self._clean_dependency_string,
+                "_desired_unit": self._desired_unit,
             }
             for dependency in self._dependency_map.values():
                 dependency._detach_observer(self)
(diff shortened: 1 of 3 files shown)


src/easyscience/variable/parameter.py
```diff
@@ -158,11 +164,19 @@ def _update(self) -> None:
             )  # noqa: E501
             self._min.unit = temporary_parameter.unit
             self._max.unit = temporary_parameter.unit
+
+            if self._desired_unit is not None:
+                try:
+                    self._convert_unit(self._desired_unit)
+                except Exception as e:
+                    raise UnitError(f'Failed to convert unit from {temporary_parameter.unit} to {self._desired_unit}: {e}')
```
rozyczko: Should we completely bail out on bad unit conversion? Maybe return the original value with the original unit and warn users of the failure of the unit conversion.

henrikjacobsenfys: Hmm, good question. I tend to think yes, since something must have gone wrong if the user expects a unit that is incompatible with the calculated unit.

src/easyscience/variable/parameter.py
```diff
@@ -470,6 +489,18 @@ def convert_unit(self, unit_str: str) -> None:
...
+    def set_desired_unit(self, unit_str: str) -> None:
+        """
+        Set the desired unit for a dependent Parameter. This will convert the parameter to the desired unit.
+
+        :param unit_str: The desired unit as a string.
+        """
+
+        if self._independent:
+            raise AttributeError('This is an independent parameter, desired unit can only be set for dependent parameters.')
+        self._desired_unit = unit_str
```
damskii9992, line 530: This is missing a check on the type of `unit_str`

src/easyscience/variable/parameter.py
damskii9992: I would move this check down to the `make_dependent_on` method and do a `_revert_dependency` if it fails.
Like with all our other checks. The update method should have no checks.
The `dependency_result` can be converted to try it out.

damskii9992 (changes requested):
I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.

src/easyscience/variable/parameter.py
henrikjacobsenfys: I moved the check. Update only checks `if self._desired_unit is not None`

henrikjacobsenfys:
> I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.

The argument 'unit' is already used at least in tests, e.g. the following, so I'm hesitant to change it. Thoughts?
```
    def test_dependent_parameter_serialization(self, clear_global_map):
        """Test serialization of parameters with dependencies."""
        # Create independent parameter
        a = Parameter(name="a", value=2.0, unit="m", min=0, max=10)
        
        # Create dependent parameter
        b = Parameter.from_dependency(
            name="b",
            dependency_expression="2 * a",
            dependency_map={"a": a},
            unit="m"
        )

**Comment by damskii9992** [comment]
> > I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.
> 
> The argument 'unit' is already used at least in tests, e.g. the following, so I'm hesitant to change it. Thoughts?
> 
> ```
>     def test_dependent_parameter_serialization(self, clear_global_map):
>         """Test serialization of parameters with dependencies."""
>         # Create independent parameter
>         a = Parameter(name="a", value=2.0, unit="m", min=0, max=10)
>         
>         # Create dependent parameter
>         b = Parameter.from_dependency(
>             name="b",
>             dependency_expression="2 * a",
>             dependency_map={"a": a},
>             unit="m"
>         )
> ```

Why does it matter that it is already used in tests? O.o That would literally be the case for any name changes we do XD

**Comment by henrikjacobsenfys** [comment]
> > > I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.
> > 
> > 
> > The argument 'unit' is already used at least in tests, e.g. the following, so I'm hesitant to change it. Thoughts?
> > ```
> >     def test_dependent_parameter_serialization(self, clear_global_map):
> >         """Test serialization of parameters with dependencies."""
> >         # Create independent parameter
> >         a = Parameter(name="a", value=2.0, unit="m", min=0, max=10)
> >         
> >         # Create dependent parameter
> >         b = Parameter.from_dependency(
> >             name="b",
> >             dependency_expression="2 * a",
> >             dependency_map={"a": a},
> >             unit="m"
> >         )
> > ```
> 
> Why does it matter that it is already used in tests? O.o That would literally be the case for any name changes we do XD

Well it might not matter, it's just that that implies that people might rely on this behavior. But maybe we shouldn't allow a unit in `from_dependency` then? Since the only reason people would give a unit is because they expect the `Parameter` to have that unit, no?

Edit: the point is that `from_dependency` has always accepted `unit` as an argument, until this PR it just didn't do anything. Now it (tries) to set the unit of the dependent `Parameter` to be `unit`.

**Comment by damskii9992** [comment]
> > > > I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.
> > > 
> > > 
> > > The argument 'unit' is already used at least in tests, e.g. the following, so I'm hesitant to change it. Thoughts?
> > > ```
> > >     def test_dependent_parameter_serialization(self, clear_global_map):
> > >         """Test serialization of parameters with dependencies."""
> > >         # Create independent parameter
> > >         a = Parameter(name="a", value=2.0, unit="m", min=0, max=10)
> > >         
> > >         # Create dependent parameter
> > >         b = Parameter.from_dependency(
> > >             name="b",
> > >             dependency_expression="2 * a",
> > >             dependency_map={"a": a},
> > >             unit="m"
> > >         )
> > > ```
> > 
> > 
> > Why does it matter that it is already used in tests? O.o That would literally be the case for any name changes we do XD
> 
> Well it might not matter, it's just that that implies that people might rely on this behavior. But maybe we shouldn't allow a unit in `from_dependency` then? Since the only reason people would give a unit is because they expect the `Parameter` to have that unit, no?
> 
> Edit: the point is that `from_dependency` has always accepted `unit` as an argument, until this PR it just didn't do anything. Now it (tries) to set the unit of the dependent `Parameter` to be `unit`.

Ahh, in this case it is simply the test that is wrong. The "unit" is a kwarg passed to the Parameter constructor, but it gets overwritten so it is a redundant argument.

File: src/easyscience/variable/parameter.py [hunk 3]
```diff
@@ -289,6 +305,13 @@ def make_dependent_on(self, dependency_expression: str, dependency_map: Optional
             raise error
         # Update the parameter with the dependency result
         self._fixed = False
+
+        if self._desired_unit is not None:
+            try:
+                dependency_result._convert_unit(self._desired_unit)
+            except Exception as e:
+                raise UnitError(f'Failed to convert unit from {dependency_result.unit} to {self._desired_unit}: {e}')
+
```
rozyczko, line 328: When the unit conversion fails, the code raises a `UnitError` but does not call `_revert_dependency()` first, leaving the parameter in an inconsistent state if unit conversion fails.
This is inconsistent with all other validation checks in `make_dependent_on()` that properly revert the dependency state before raising.

src/easyscience/variable/parameter.py
```diff
@@ -470,6 +494,21 @@ def convert_unit(self, unit_str: str) -> None:
         """
         self._convert_unit(unit_str)
 
+    def set_desired_unit(self, unit_str: str) -> None:
+        """
+        Set the desired unit for a dependent Parameter. This will convert the parameter to the desired unit.
```
rozyczko: `make_dependent_on` and `from_dependency` methods accept `unit: str | sc.Unit | None` but this only takes `str` - is this expected?

src/easyscience/variable/parameter.py
```diff
@@ -158,11 +164,17 @@ def _update(self) -> None:
...
             self._notify_observers()
         else:
             warnings.warn('This parameter is not dependent. It cannot be updated.')
 
-    def make_dependent_on(self, dependency_expression: str, dependency_map: Optional[dict] = None) -> None:
+    def make_dependent_on(
+        self, dependency_expression: str, dependency_map: Optional[dict] = None, unit: str | sc.Unit | None = None
+    ) -> None:
         """
         Make this parameter dependent on another parameter. This will overwrite the current value, unit, variance, min and max.
```
rozyczko: There's no validation on `unit` type. A user can pass `unit=1234` and this will get accepted.

src/easyscience/variable/parameter.py
rozyczko, line 236: You forgot to add `_desired_unit` to previous dependency

src/easyscience/variable/parameter.py
```diff
@@ -158,11 +164,17 @@ def _update(self) -> None:
             )  # noqa: E501
             self._min.unit = temporary_parameter.unit
             self._max.unit = temporary_parameter.unit
+
+            if self._desired_unit is not None:
+                self._convert_unit(self._desired_unit)
+
```
rozyczko, line 174: `_convert_unit` doesn't catch any exceptions and we aren't checking anything here. Maybe add some exception handling so it doesn't bubble up potentially leaving the parameter in a weird state?

tests/unit_tests/variable/test_parameter.py
```diff
@@ -136,6 +136,49 @@ def test_make_dependent_on(self, normal_parameter: Parameter):
         normal_parameter.value == 4
         self.compare_parameters(normal_parameter, 2*independent_parameter)
 
+
```
rozyczko: There's no test verifying that `_desired_unit` is properly handled when calling `make_dependent_on` on an already-dependent parameter (i.e., overwriting an existing dependency).

rozyczko (changes requested):
A few issues noted.

src/easyscience/variable/parameter.py
damskii9992: This is on purpose. To keep the `_update` method as fast as possible, all the checks are made in the `make_dependent_on` method. If this method succeeds, then the `_update` will also always succeed.
This is also how we did it for the other updates :)

henrikjacobsenfys:
> > > > > I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.
> > > > 
> > > > 
> > > > The argument 'unit' is already used at least in tests, e.g. the following, so I'm hesitant to change it. Thoughts?
> > > > ```
> > > >     def test_dependent_parameter_serialization(self, clear_global_map):
> > > >         """Test serialization of parameters with dependencies."""
> > > >         # Create independent parameter
> > > >         a = Parameter(name="a", value=2.0, unit="m", min=0, max=10)
> > > >         
> > > >         # Create dependent parameter
> > > >         b = Parameter.from_dependency(
> > > >             name="b",
> > > >             dependency_expression="2 * a",
> > > >             dependency_map={"a": a},
> > > >             unit="m"
> > > >         )
> > > > ```
> > > 
> > > 
> > > Why does it matter that it is already used in tests? O.o That would literally be the case for any name changes we do XD
> > 
> > 
> > Well it might not matter, it's just that that implies that people might rely on this behavior. But maybe we shouldn't allow a unit in `from_dependency` then? Since the only reason people would give a unit is because they expect the `Parameter` to have that unit, no?
> > Edit: the point is that `from_dependency` has always accepted `unit` as an argument, until this PR it just didn't do anything. Now it (tries) to set the unit of the dependent `Parameter` to be `unit`.
> 
> Ahh, in this case it is simply the test that is wrong. The "unit" is a kwarg passed to the Parameter constructor, but it gets overwritten so it is a redundant argument.

Because all other `Parameter` constructors accept `unit` and make it so that the `Parameter` has the chosen `unit` I'm in favor of keeping the name here as well. It makes sense to me that the unit of the `Parameter` is `unit`, also if it's dependent.

If you/the the team prefer a different name (`desired_unit` or `preferred_unit` ?) should we then still allow `unit` as a kwarg? I feel like that would potentially be confusing to users, since they'd get no error or warning with this code:
```
b = Parameter.from_dependency(
             name="b",
             dependency_expression="2 * a",
             dependency_map={"a": a},
             unit="m",
             desired_unit = "cm"
         )
```

I'm also curious to hear what @rozyczko  and @AndrewSazonov think about this :)

src/easyscience/variable/parameter.py
henrikjacobsenfys: I didn't realise `dependency_result` changed self. Adding `_revert_dependency()` before raising the error.

src/easyscience/variable/parameter.py
henrikjacobsenfys: Just an oversight.

src/easyscience/variable/parameter.py
henrikjacobsenfys: Added validation.

tests/unit_tests/variable/test_parameter.py
henrikjacobsenfys: Good point! Expanded the `test_dependent_parameter_make_dependent_on_with_desired_unit` test

damskii9992:
> > > > > > I wouldn't simply call this argument "unit", it is too vague and can be understood as it can be anything. Maybe "unit_format"? Or something similar.
> > > > > 
> > > > > 
> > > > > The argument 'unit' is already used at least in tests, e.g. the following, so I'm hesitant to change it. Thoughts?
> > > > > ```
> > > > >     def test_dependent_parameter_serialization(self, clear_global_map):
> > > > >         """Test serialization of parameters with dependencies."""
> > > > >         # Create independent parameter
> > > > >         a = Parameter(name="a", value=2.0, unit="m", min=0, max=10)
> > > > >         
> > > > >         # Create dependent parameter
> > > > >         b = Parameter.from_dependency(
> > > > >             name="b",
> > > > >             dependency_expression="2 * a",
> > > > >             dependency_map={"a": a},
> > > > >             unit="m"
> > > > >         )
> > > > > ```
> > > > 
> > > > 
> > > > Why does it matter that it is already used in tests? O.o That would literally be the case for any name changes we do XD
> > > 
> > > 
> > > Well it might not matter, it's just that that implies that people might rely on this behavior. But maybe we shouldn't allow a unit in `from_dependency` then? Since the only reason people would give a unit is because they expect the `Parameter` to have that unit, no?
> > > Edit: the point is that `from_dependency` has always accepted `unit` as an argument, until this PR it just didn't do anything. Now it (tries) to set the unit of the dependent `Parameter` to be `unit`.
> > 
> > 
> > Ahh, in this case it is simply the test that is wrong. The "unit" is a kwarg passed to the Parameter constructor, but it gets overwritten so it is a redundant argument.
> 
> Because all other `Parameter` constructors accept `unit` and make it so that the `Parameter` has the chosen `unit` I'm in favor of keeping the name here as well. It makes sense to me that the unit of the `Parameter` is `unit`, also if it's dependent.
> 
> If you/the the team prefer a different name (`desired_unit` or `preferred_unit` ?) should we then still allow `unit` as a kwarg? I feel like that would potentially be confusing to users, since they'd get no error or warning with this code:
> 
> ```
> b = Parameter.from_dependency(
>              name="b",
>              dependency_expression="2 * a",
>              dependency_map={"a": a},
>              unit="m",
>              desired_unit = "cm"
>          )
> ```
> 
> I'm also curious to hear what @rozyczko and @AndrewSazonov think about this :)

The problem is that the `unit` that the user passes here will not be any unit like it can be for any other `Parameter` constructor.
In the regular `Parameter` constructor, the `unit` is the actual unit of the parameter, it gets set to this value.
When making a dependent parameter, the user cannot set the value of the unit, as it is inferred from the dependency expression (like the value, min/max and error). So if the user want e.g. "Meter" and the dependency expression returns a "Second" unit, the unit cannot be converted and it would result in an error, completely different from the regular `unit` setters.

We pass kwargs to the `Parameter` constructor for stuff like `display_name`, `url`, `unique_name` etc. For all the values inferred by the dependency evaluation, whatever the user provides gets overwritten. The user can also provide a "value", but just like the "unit", this kwarg simply gets passed on and overwritten immediately. This is just a side-effect of allowing arbitrary kwargs to be passed to the underlying `Parameter` constructor.

And this is why I want this "desired_unit" or "preffered_unit" to be named differently, because it is different from the regular "unit".

src/easyscience/variable/parameter.py
henrikjacobsenfys: Added

rozyczko:
Why is this PR so large with seemingly unrelated formatting changes?
The exisiting code already passes the stringent ruff check and ruff format tests so why update?

henrikjacobsenfys:
> Why is this PR so large with seemingly unrelated formatting changes? The exisiting code already passes the stringent ruff check and ruff format tests so why update?

I don't know! I ran "pixi run -e dev format" just now, but looks like there's still a lot of formatting happening. I don't know if I can do more :/

damskii9992 (approved):
Approved

rozyczko (approved):
LGTM+

PR #193, opened 2026-02-05
Add initial EasyList class

damskii9992:
Here is my initial thoughts on how the generic EasyList could look like. I haven't added tests yet, but please have a look.

### DIFF
--- a/src/easyscience/base_classes/easy_list.py
+++ b/src/easyscience/base_classes/easy_list.py
@@ -111,20 +111,24 @@ class EasyList(NewBase, MutableSequence[ProtectedType_]):
         if isinstance(idx, int):
             if not isinstance(value, tuple(self._protected_types)):
                 raise TypeError(f'Items must be one of {self._protected_types}, got {type(value)}')
-            if value in self:
+            if value is not self._data[idx] and value in self:
                 warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
                 return
             self._data[idx] = value
         elif isinstance(idx, slice):
             if not isinstance(value, Iterable):
                 raise TypeError('Value must be an iterable for slice assignment')
-            for v in value:
+            replaced = self._data[idx]
+            new_values = list(value)
+            if len(new_values) != len(replaced):
+                raise ValueError('Length of new values must match the length of the slice being replaced')
+            for i, v in enumerate(new_values):
                 if not isinstance(v, tuple(self._protected_types)):
                     raise TypeError(f'Items must be one of {self._protected_types}, got {type(v)}')
-                if v in self:
-                    warnings.warn(f'Item with unique name "{self._get_key(v)}" already in EasyList, it will be ignored')
-                    return
-            self._data[idx] = list(value)  # type: ignore[arg-type]
+                if v in self and replaced[i] is not v:
+                    warnings.warn(f'Item with unique name "{v.unique_name}" already in EasyList, it will be ignored')
+                    new_values[i] = replaced[i]  # Keep the original value if the new one is a duplicate
+            self._data[idx] = new_values
         else:
             raise TypeError('Index must be an int or slice')
 
--- a/src/easyscience/base_classes/easy_list.py
+++ b/src/easyscience/base_classes/easy_list.py
@@ -90,7 +90,7 @@ class EasyList(NewBase, MutableSequence[ProtectedType_]):
         elif isinstance(idx, slice):
             return self.__class__(self._data[idx], protected_types=self._protected_types)
         elif isinstance(idx, str):
-            element = next((r for r in self._data if r.unique_name == idx), None)
+            element = next((r for r in self._data if self._get_key(r) == idx), None)
             if element is not None:
                 return element
             raise KeyError(f'No item with unique name "{idx}" found')
@@ -133,7 +133,7 @@ class EasyList(NewBase, MutableSequence[ProtectedType_]):
             del self._data[idx]
         elif isinstance(idx, str):
             for i, item in enumerate(self._data):
-                if item.unique_name == idx:
+                if self._get_key(item) == idx:
                     del self._data[i]
                     return
             raise KeyError(f'No item with unique name "{idx}" found')
@@ -158,6 +158,17 @@ class EasyList(NewBase, MutableSequence[ProtectedType_]):
 
         self._data.insert(index, value)
 
+    def _get_key(self, object, attribute='unique_name') -> str:
+        """
+        Get the unique name of an object.
+        Can be overridden to use a different attribute as the key.
+        :param object: Object to get the key for
+        :param attribute: Attribute to use as the key (default is 'unique_name')
+        :return: The key of the object
+        :rtype: str
+        """
+        return getattr(object, attribute)
+
     # Overwriting methods
 
     def sort(self, mapping: Callable[[ProtectedType_], Any], reverse: bool = False) -> None:
@@ -177,13 +188,13 @@ class EasyList(NewBase, MutableSequence[ProtectedType_]):
 
     def __contains__(self, item: ProtectedType_ | str) -> bool:
         if isinstance(item, str):
-            return any(r.unique_name == item for r in self._data)
+            return any(self._get_key(r) == item for r in self._data)
         return item in self._data
 
     def index(self, value: ProtectedType_ | str, start: int = 0, stop: int = ...) -> int:
         if isinstance(value, str):
             for i in range(start, min(stop, len(self._data))):
-                if self._data[i].unique_name == value:
+                if self._get_key(self._data[i]) == value:
                     return i
             raise ValueError(f'{value} is not in EasyList')
         return self._data.index(value, start, stop)
@@ -209,7 +220,7 @@ class EasyList(NewBase, MutableSequence[ProtectedType_]):
             return self._data.pop(index)
         elif isinstance(index, str):
             for i, item in enumerate(self._data):
-                if item.unique_name == index:
+                if self._get_key(item) == index:
                     return self._data.pop(i)
             raise KeyError(f'No item with unique name "{index}" found')
         else:
--- a/src/easyscience/base_classes/easy_list.py
+++ b/src/easyscience/base_classes/easy_list.py
@@ -1,0 +1,267 @@
+#  SPDX-FileCopyrightText: 2025 EasyScience contributors  <core@easyscience.software>
+#  SPDX-License-Identifier: BSD-3-Clause
+#  © 2021-2025 Contributors to the EasyScience project <https://github.com/easyScience/EasyScience
+
+from __future__ import annotations
+
+from collections.abc import MutableSequence
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Callable
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Type
+from typing import TypeVar
+from typing import overload
+
+from pyparsing import Dict
+
+from easyscience.io.serializer_base import SerializerBase
+
+from .new_base import NewBase
+
+if TYPE_CHECKING:
+    pass
+
+ProtectedType_ = TypeVar('ProtectedType', bound=NewBase)
+
+
+class EasyList(NewBase, MutableSequence[ProtectedType_]):
+    def __init__(
+        self,
+        *args: ProtectedType_ | list[ProtectedType_],
+        protected_types: list[Type[NewBase]] | Type[NewBase] | None = None,
+        unique_name: Optional[str] = None,
+        display_name: Optional[str] = None,
+        **kwargs: Any,
+    ):
+        """
+        Initialize the EasyList.
+        :param args: Initial items to add to the list
+        :param protected_types: Types that are allowed in the list. Can be a single NewBase subclass or a list of them.
+                                If None, defaults to [NewBase].
+        :param unique_name: Optional unique name for the list
+        :param display_name: Optional display name for the list
+        """
+        super().__init__(unique_name=unique_name, display_name=display_name)
+        if protected_types is None:
+            self._protected_types = [NewBase]
+        elif isinstance(protected_types, type) and issubclass(protected_types, NewBase):
+            self._protected_types = [protected_types]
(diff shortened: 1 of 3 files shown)


src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def __iter__(self) -> Any:
+        return iter(self._data)
+
+    def __contains__(self, item: ProtectedType_ | str) -> bool:
+        if isinstance(item, str):
+            return any(r.unique_name == item for r in self._data)
+        return item in self._data
+
+    def index(self, value: ProtectedType_ | str, start: int = 0, stop: int = ...) -> int:
```
rozyczko: I think this will throw TypeError in `min(stop...)` if `stop` is left as `...` (ellipsis can't be compared to int)

You might need to change it to `stop: int | None = None` instead

rozyczko (commented):
- I understand that the lack of graph integration is a feature? We let the list components do the edge/vertex tracking?

- Good type safety but not sure we should allow duplicates. My ModelCollection just skips duplicates which might be even worse.

- Where is the `data` property? Do we need it?

- Where is the aggregate `get_all_variables`? Just run over all children and return that aggregate.

henrikjacobsenfys:
I can find no problems with the code, but it's also using some methods that are new to me, so I'm not an expert.

As discussed elsewhere, I'll want this exact class, but using something like `name` instead of `unique_name`.

From what I understand, we could make that somewhat straightforward by introducing something like 

```
def _get_key(self, item: ProtectedType_) -> str:
    return item.unique_name
```
and replace calls to `item.unique_name`  with `self._get_key(item)`.
That way I can inherit from this class and override `_get_key` (and write my own validation to ensure the `name` is unique within the specific list).

This seems to me to be cleaner than maintaining a copy of EasyList in EasyDynamics.

I approve the name BTW.

src/easyscience/base_classes/easy_list.py
damskii9992: Yeah, I was a little fast here with the Copilot auto-complete. I'm surprised it suggested the Ellipsis as the default though. I guess it must be used extensively in Numpy and Scipy which is where Copilot got it from :/

damskii9992:
For documentation:
> * I understand that the lack of graph integration is a feature? We let the list components do the edge/vertex tracking?

Yes, this is another step towards simplifying our `global_object `and removing our use of the graph to eventually remove it completely. The end-goal is for the `global_object `to only have a weakref dictionary of key-value pairs of `unique_names`. I.e. get rid of all the other complicated and unused functionality.
If you have a good use-case for keeping the graph functionality, please write it down so we can asses our actual needs and what we can do about it.

> * Good type safety but not sure we should allow duplicates. My ModelCollection just skips duplicates which might be even worse.

Do you mean duplicates of the same object in the list? I guess we can safeguard against that by adding a check in the __setitem__ and append methods. But that will slow down setting and appending, but I guess we will never use `EasyList `for huge lists, so maybe the slow down is okay? Is there any use-case where we might want duplicates of an item in a list?

> * Where is the `data` property? Do we need it?

The 'data' property is an internal only property. It is also used for serialization. I did consider making it public, but then it would be weird that you have 2 ways of setting the lists data, i.e.:

```python
EasyList(1,2,3)

EasyList(data=[1,2,3])

EasyList(1,2,3, data=[4,5,6])
```
And what would we do in the last case when both are provided? I felt it was simpler/cleaner to simply make `data` an internal and serialization only attribute.

> * Where is the aggregate `get_all_variables`? Just run over all children and return that aggregate.
That would be for a child-class, lets call it `ModelCollection` which inherits from this and where the `protected_types `has to inherit from `ModelBase` rather than `NewBase` like in this one. Since all elements would be instances of `ModelBase`, we would then know that they all have their own `get_all_variables` methods.

seventil:
LGTM, upvote for the (optional) uniqueness checks. 
> but using something like name instead of unique_name.
we really should write down all the functional requirements for names

rozyczko:
remember to fix ruff format...

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+        Append an item to the end of the collection.
+
+        :param value: Item to append
+        """
+        if not isinstance(value, tuple(self._protected_types)):
+            raise TypeError(f'Items must be one of {self._protected_types}, got {type(value)}')
+        if value in self:
+            warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
+            return
+        self._data.append(value)
```
rozyczko: `MutableSequence` append is calling `insert`. No need to override this method at all.
You are already doing all the relevant tests in `insert`.

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+        :param mapping: Mapping function to sort by
+        :param reverse: Whether to reverse the sort
+        """
+        self._data.sort(key=mapping, reverse=reverse)  # type: ignore[arg-type]
+
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__} of length {len(self)} of type(s) {self._protected_types}'
+
+    def __iter__(self) -> Any:
+        return iter(self._data)
```
rozyczko: MutableSequence provides a default that iterates via __getitem__(i) for i = 0, 1, 2, .... Your override (iter(self._data)) is functionally identical — just a micro-optimization skipping the index dispatch.

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+from typing import Any
+from typing import Callable
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Type
+from typing import TypeVar
+from typing import overload
+
+from pyparsing import Dict
```
rozyczko: Why `pyparsing` not just the standard `typing.Dict`??

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+            self._data[idx] = value
+        elif isinstance(idx, slice):
+            if not isinstance(value, Iterable):
+                raise TypeError('Value must be an iterable for slice assignment')
+            for v in value:
+                if not isinstance(v, tuple(self._protected_types)):
+                    raise TypeError(f'Items must be one of {self._protected_types}, got {type(v)}')
+                if v in self:
+                    warnings.warn(f'Item with unique name "{self._get_key(v)}" already in EasyList, it will be ignored')
+                    return
```
rozyczko: This is **bad**.

`reverse()` in MutableSequence swaps elements via `__setitem__`:
```python
# MutableSequence.reverse() effectively does:
for i in range(n // 2):
    self[i], self[n-i-1] = self[n-i-1], self[i]
```
But your __setitem__ uniqueness guard rejects items that are already in the list. So `reverse()` will just warn and return.

Look:

```
>>> from easyscience.base_classes import NewBase, EasyList
>>> b1 = NewBase()
>>> b2 = NewBase()
>>> a = EasyList([b1, b2])
>>> list(a.reverse())
C:\projects\easy\ERA\corelib\src\easyscience\base_classes\easy_list.py:117: UserWarning: Item with unique name "NewBase_1" already in EasyList, it will be ignored
  warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
C:\projects\easy\ERA\corelib\src\easyscience\base_classes\easy_list.py:117: UserWarning: Item with unique name "NewBase_0" already in EasyList, it will be ignored
  warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```

This potentially affects other methods which use direct assignment with `__setitem__`

rozyczko: Also, 
```python
    warnings.warn(f'Item with unique name "{self._get_key(v)}" already in EasyList, it will be ignored')
    return
```

`return` should be changed to `continue`, otherwise you will return without actually setting any item.

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+        :param idx: Index to set
+        :param value: New value
+        """
+        if isinstance(idx, int):
+            if not isinstance(value, tuple(self._protected_types)):
+                raise TypeError(f'Items must be one of {self._protected_types}, got {type(value)}')
+            if value in self:
+                warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
+                return
+            self._data[idx] = value
```
rozyczko, line 117: in-place replacement will fail.

```python
>>> from easyscience.base_classes import NewBase, EasyList
>>> a1 = NewBase()
>>> a2 = NewBase()
>>> e = EasyList([a1, a2])
>>> e[0] = e[0]
C:\projects\easy\ERA\corelib\src\easyscience\base_classes\easy_list.py:117: UserWarning: Item with unique name "NewBase_0" already in EasyList, it will be ignored
  warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
>>> e[0] = e[1]
C:\projects\easy\ERA\corelib\src\easyscience\base_classes\easy_list.py:117: UserWarning: Item with unique name "NewBase_1" already in EasyList, it will be ignored
  warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
```

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+
+        :param obj_dict: dictionary containing the serialized contents (from `SerializerDict`) of an EasyScience object
+        :return: Reformed EasyScience object
+        """
+        if not SerializerBase._is_serialized_easyscience_object(obj_dict):
+            raise ValueError('Input must be a dictionary representing an EasyScience EasyList object.')
+        if obj_dict['@class'] == cls.__name__:
+            if 'protected_types' in obj_dict:
+                protected_types = obj_dict.pop('protected_types')
+                for i, type_dict in enumerate(protected_types):
```
rozyczko: you are mutating the input `obj_dict` here. This is BAD.
Use `obj_dict.get(...)` or create a copy

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+            raise TypeError('Index must be an integer')
+        elif not isinstance(value, tuple(self._protected_types)):
+            raise TypeError(f'Items must be one of {self._protected_types}, got {type(value)}')
+        if value in self:
+            warnings.warn(f'Item with unique name "{self._get_key(value)}" already in EasyList, it will be ignored')
+            return
+        self._data.insert(index, value)
+
+    def _get_key(self, object) -> str:
+        """
```
rozyczko: you don't want to shadow the builtin `object`... rename it to something like `obj`

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+from typing import overload
+
+from pyparsing import Dict
+
+from easyscience.io.serializer_base import SerializerBase
+
+from .new_base import NewBase
+
+if TYPE_CHECKING:
+    pass
```
rozyczko: remove

src/easyscience/base_classes/easy_list.py
```diff
@@ -0,0 +1,289 @@
...
+        :param object: Object to get the key for
+        :return: The key of the object
+        :rtype: str
+        """
+        return object.unique_name
+
+    # Overwriting methods
+
+    def sort(self, mapping: Callable[[ProtectedType_], Any], reverse: bool = False) -> None:
+        """
```
rozyczko: very non-standard signature. `list.sort()` uses `key`, not `maping`.

```python
items = [3, 1, 2]
items.sort(key=lambda x: x)  # this is what I would expect with EasyList

# EasyList:
el = EasyList(a3, a1, a2)
el.sort(key=lambda x: x.unique_name) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: EasyList.sort() got an unexpected keyword argument 'key'
```

What works is an unintuitive

```python
el.sort(mapping=lambda x: x.unique_name) 
```

rozyczko (changes requested):
Some serious issues found. Please consider addressing them.

src/easyscience/base_classes/easy_list.py
damskii9992: I fixed the `reverse()` issue by overwriting `__reversed__`, as far as I am aware, there is no other methods which use direct assignment with `__setitem__` (I looked through the list of inherited methods in the Collections.abc), if you know or can think of any, please do tell.

damskii9992: Do we really want it to still set the other elements? So essentially, when setting with a slice, we overwrite all the elements EXCEPT the faulty one?
I think it's much better to no set any of them at all . . .

src/easyscience/base_classes/easy_list.py
damskii9992: Good find. Not that we really expect the dictionaries to be re-used, but it is bad practice to edit input. I added a deepcopy operation.

damskii9992: And a test. Of course.

src/easyscience/base_classes/easy_list.py
damskii9992: This was copied over from your original ModelCollection which probably copied it over from the original BaseCollection. I assume Simon didn't like the original syntax or some sort? I agree though, better to keep it in line with normal python list syntax.

src/easyscience/base_classes/easy_list.py
damskii9992: I fixed this issue with my latest commit. The implementation might not be fast (it's just a for-loop), but the `EasyList` was never meant to be a super fast class which holds super long lists. If we need it to be faster in the future, then we can re-asses the implementation.

src/easyscience/base_classes/easy_list.py
rozyczko: I don't think there are any more.

rozyczko (approved):
Looks good now.

PR #196, opened 2026-02-09
Attempt at creating monotonic unique_id counters

rozyczko:
To avoid the issue with GC messing up with unique names during internal copy operations, let's just have a simple(r) monotonically increasing counters for `unique_name`s.

This addresses https://github.com/easyscience/corelib/issues/195 and the original issue in [dynamics-lib](https://ess-eric.slack.com/archives/C06HSEXN7K2/p1770528590198159)


rozyczko:
Closing since the actual fix should be done in dynamics-lib

PR #197, opened 2026-02-10
Reordered operations in Map.add_vertex

seventil:
Fix is made to avoid hitting a race condition due to weakref.finalize() being called. It seems that .finalize temporarily deletes references of the object passed to it from dicts, therefore trying to get it from __type_dict or _store might randomly raise a KeyError.

This PR fixes failing tests from https://github.com/easyscience/dynamics-lib/pull/96

### DIFF
--- a/src/easyscience/global_object/map.py
+++ b/src/easyscience/global_object/map.py
@@ -143,10 +143,13 @@ class Map:
         # but the finalizer hasn't run yet
         if name in self.__type_dict:
             del self.__type_dict[name]
+
         self._store[name] = obj
-        self.__type_dict[name] = _EntryList()  # Add objects type to the list of types
-        self.__type_dict[name].finalizer = weakref.finalize(self._store[name], self.prune, name)
-        self.__type_dict[name].type = obj_type
+
+        entry_list = _EntryList()
+        entry_list.finalizer = weakref.finalize(obj, self.prune, name)
+        entry_list.type = obj_type
+        self.__type_dict[name] = entry_list  # Add objects type to the list of types
 
     def add_edge(self, start_obj: object, end_obj: object):
         if start_obj.unique_name in self.__type_dict:


rozyczko:
How is this fixing the issue? Can you explain what those lines are changing?

seventil:
> How is this fixing the issue? Can you explain what those lines are changing?

Just for reference, [the traceback that lead to the creation of this PR](https://github.com/easyscience/dynamics-lib/actions/runs/21703278855/job/62588257853?pr=96)
```
        self.__type_dict[name] = _EntryList()  # Add objects type to the list of types
>       self.__type_dict[name].finalizer = weakref.finalize(self._store[name], self.prune, name)
        ^^^^^^^^^^^^^^^^^^^^^^
E       KeyError: 'Parameter_338'
```

Weird thing is that here we see sequential `__setitem__` call and `__getitem__` call on the same dict resulting in KeyError, which just doesn't make any sense.

Here is my assumption of what is happening:
when wekref.finalize() is being called with some object passed as an argument, the dicts that hold a reference to such object are being renewed -> key-val pairs dissapear and then return back to the dicts. In some cases, trying to get an item from a dict while the dicts are being 'renewed' results in a KeyError. This is just my observation from try runs, I don't know how `finalize` actually works underneath the python API.

Basically, the line `self.__type_dict[name].finalizer = weakref.finalize(self._store[name], self.prune, name)` makes unnecesary calls to `self.__type_dict.__getitem__` and to `self._store.__getitem__`.
If removing unnecessary calls also happens to solve the issue with "disappearing" keys, I think it's a worthwile change.
It doesn't change the logic of the method, but rather just semantics, so why not try?

I tested this change in EasyDynamics, which fixed the failing tests from https://github.com/easyscience/dynamics-lib/pull/96

I didn't investigate why tests fail in this branch, but I expect that `more_weakref_fixes` hold the fixes for the issue I have here.

seventil:
@rozyczko could you please merge this into your branch if it's ok?

PR #217, opened 2026-02-18
Changed the finalizer function for __type_dict

seventil:
This is a temporary quickfix, until __type_dict is pruned from the codebase. The need for this change is due to serialization of baseobjects creating temprorary objects. When the temp objects are finally GC'd, their name in the _store dict might be already taken by an actual (not a temporary) Parameter. GC then calls the the __type_dict's finalizer. In the old solution it was the prune method, which removes the key from __type_dict as well as _store (which might be an actual Parameter by that time). The new solution provides a method to delete keys only from the __type_dict, since removing keys from _store might affect actual Parameters and since _store is a weakref dict anyway.

### DIFF
--- a/src/easyscience/global_object/map.py
+++ b/src/easyscience/global_object/map.py
@@ -169,7 +169,7 @@ class Map:
         self._store[name] = obj
 
         entry_list = _EntryList()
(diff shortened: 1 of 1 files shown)


damskii9992 (approved):
Yes, lets get this fucker merged!

PR #226, opened 2026-03-09
new collection based on easy_list

rozyczko:
CollectionBase derived from EasyList

- Derives from `EasyList` rather than re-implementing `MutableSequence` from scratch.
- Removes all functionality now owned by `EasyList` (sequence operations, type protection, deduplication, serialization).
- Assumes legacy collection-specific concerns such as graph edges, interface propagation,
    and `NotarizedDict` have already been removed before this class is introduced.

Inheritance Hierarchy

```
NewBase
  └── EasyList[T]                   ← full MutableSequence, serialization, type protection
        └── CollectionBase[T]       ← adds aggregation methods only
```

content:

1. Parameter / Variable Aggregation

get_all_variables(self) 
get_all_parameters(self) 
get_parameters(self)
get_fittable_parameters(self) 
get_free_parameters(self) 
get_fit_parameters(self) 

These mirror the methods on `ModelBase`, but aggregate across the **list of items** rather than
across the attribute names of a single object.

2.  `data` Property

3. Legacy Serialization (`_convert_to_dict`)
Support for the legacy encoder-based serialization path.
Serializes all items in _data under the 'data' key, consistent with the old CollectionBase._convert_to_dict signature.

### DIFF
--- a/src/easyscience/base_classes/__init__.py
+++ b/src/easyscience/base_classes/__init__.py
@@ -1,6 +1,6 @@
 from .based_base import BasedBase
-from .collection_base_easylist import CollectionBase as CollectionBaseEasyList
 from .collection_base import CollectionBase
+from .collection_base_easylist import CollectionBase as CollectionBaseEasyList
 from .easy_list import EasyList
 from .model_base import ModelBase
 from .new_base import NewBase
--- a/src/easyscience/base_classes/__init__.py
+++ b/src/easyscience/base_classes/__init__.py
@@ -1,8 +1,9 @@
 from .based_base import BasedBase
+from .collection_base_easylist import CollectionBase as CollectionBaseEasyList
 from .collection_base import CollectionBase
 from .easy_list import EasyList
 from .model_base import ModelBase
 from .new_base import NewBase
 from .obj_base import ObjBase
 
-__all__ = [BasedBase, CollectionBase, ObjBase, ModelBase, NewBase, EasyList]
+__all__ = [BasedBase, CollectionBase, CollectionBaseEasyList, ObjBase, ModelBase, NewBase, EasyList]
--- a/src/easyscience/base_classes/collection_base.py
+++ b/src/easyscience/base_classes/collection_base.py
@@ -4,416 +4,249 @@
 
 from __future__ import annotations
 
-import copy
-import warnings
-from collections.abc import Iterable
-from importlib import import_module
+from collections.abc import MutableSequence
 from numbers import Number
+from typing import TYPE_CHECKING
 from typing import Any
+from typing import Callable
+from typing import List
 from typing import Optional
+from typing import Tuple
+from typing import Union
 
-from easyscience.io.serializer_base import SerializerBase
-from easyscience.io.serializer_dict import SerializerDict
+from easyscience.base_classes.new_base import NewBase
+from easyscience.global_object.undo_redo import NotarizedDict
 
 from ..variable.descriptor_base import DescriptorBase
-from ..variable.parameter import Parameter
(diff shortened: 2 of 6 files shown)


rozyczko:
Question: 
- do we need `named_items` argument and handling?
This would potentially allow callers to pass EasyScience objects as keyword arguments, which looks more natural.
E.g.

```python
cell = CollectionBase(name='Cell', length_a=length_a, length_b=length_b, length_c=length_c)
```

as opposed to when we need positional args
```python
cell = CollectionBase(length_a, length_b, length_c, name='Cell')
```

The keyword names are discarded, and only the values are kept as items.
Currently, the key names are checked against `_RESERVED_NAMED_KEYS` to prevent collisions with actual constructor parameters like `name` or `data`, and None values are silently dropped.

damskii9992:
> Question:
> 
> * do we need `named_items` argument and handling?
>   This would potentially allow callers to pass EasyScience objects as keyword arguments, which looks more natural.
>   E.g.
> 
> ```python
> cell = CollectionBase(name='Cell', length_a=length_a, length_b=length_b, length_c=length_c)
> ```
> 
> as opposed to when we need positional args
> 
> ```python
> cell = CollectionBase(length_a, length_b, length_c, name='Cell')
> ```
> 
> The keyword names are discarded, and only the values are kept as items. Currently, the key names are checked against `_RESERVED_NAMED_KEYS` to prevent collisions with actual constructor parameters like `name` or `data`, and None values are silently dropped.

I don't really see the value of allowing keyword arguments? This is a list-class, i.e. it is more natural to add elements in a list-like fashion, i.e:
```python
best_list = [model_1, model_2, model_3]
model_collection = ModelCollection(model_1, model_2, model_3)
```
The keyword arguments of the previous implementation was necessary to inject objects as attributes to the class, the keys were used for the attribute names, hence the reason to check for clashes. Since we don't want to allow parameter injection, we don't need this feature.

AndrewSazonov:
My reply will overlap with Christian’s one, but I will still add it, since I already finished writing it before I saw his comment.

When I see this example

```python
cell = CollectionBase(name='Cell', length_a=length_a, length_b=length_b, length_c=length_c)
```

I am a bit confused.

In my understanding, a `cell` object should not be a collection, because it contains different types of parameters (name, lengths, angles). I thought that `CollectionBase` is mainly for cases where we have a collection of the same type of objects, for example atom sites.

In that case, we would have something like:

```python
atom_site = AtomSite(label='Si', type_symbol='Si', fract_x=0.5, fract_y=0, fract_z=1, b_iso=1, occupancy=1)
```

Here keyword arguments are very important, because the positional version

```python
atom_site = AtomSite('Si', 'Si', 0.5, 0, 1, 1, 1)
```

is not very clear.

Then we create a collection:

```python
atom_sites = CollectionBase()
```

and append items:

```python
atom_sites.append(atom_site1)
atom_sites.append(atom_site2)
```

or maybe add them in bulk (e.g. list / numpy array).

If we create a collection like:

```python
atom_sites = CollectionBase(atom_site1, atom_site2, ...)
```

then using keyword names for items does not make sense. So, I think the named_items argument is not needed.

rozyczko:
There's still a legacy layer on top of the class, so the current Parameters/Descriptors can be added as members.

damskii9992:
251 files changed and 40 000 lines of code???? 😨

rozyczko:
Meh, merging the new core completely screwed up the branch. Need to redo the changes based on the current `develop`

rozyczko:
closing for the updated version in https://github.com/easyscience/core/pull/228

PR #228, opened 2026-03-18
Collections based on ModelBase

rozyczko:
This is the collection list as described in https://github.com/easyscience/core/pull/226
Rebased against the new `develop` branch and due to the number of files affected, made a separate PR.

For discussion look in: https://github.com/easyscience/core/pull/226

### DIFF
--- a/src/easyscience/base_classes/collection_base_easylist.py
+++ b/src/easyscience/base_classes/collection_base_easylist.py
@@ -5,45 +5,30 @@
 from __future__ import annotations
 
 import copy
-import warnings
 from collections.abc import Iterable
+from collections.abc import MutableSequence
 from importlib import import_module
 from numbers import Number
 from typing import Any
 from typing import Optional
 
 from easyscience.io.serializer_base import SerializerBase
-from easyscience.io.serializer_dict import SerializerDict
 
 from ..variable.descriptor_base import DescriptorBase
-from ..variable.parameter import Parameter
-from .based_base import BasedBase
-from .easy_list import EasyList
-from .new_base import NewBase
+from .model_base import ModelBase
 
 
-class CollectionBase(EasyList):
-    """EasyList-backed collection with a small compatibility layer for
-    migration.
+class CollectionBase(ModelBase, MutableSequence):
+    """A ModelBase-backed collection that can contain only models and
+    other collections.
 
-    The collection delegates storage and MutableSequence behavior to
-    EasyList, adding only scientific-parameter aggregation methods and a
-    thin compatibility layer for existing callers.
+    Inherits from ``ModelBase`` (identity, serialisation, parameter aggregation)
+    and ``MutableSequence`` (list-like interface).  Parameter aggregation
+    methods are overridden to recurse into the contained items rather than
+    using ``dir()`` introspection.
     """
 
-    # Base types that items in this collection must be instances of when no
-    # custom ``protected_types`` are supplied by the caller.  The validation
-    # check in ``_validate_item`` uses this list to reject plain Python objects
-    # that are not part of the EasyScience type hierarchy.
-    # ``BasedBase`` is only kept for backwards compatibility.
-    _DEFAULT_PROTECTED_TYPES = (DescriptorBase, BasedBase, NewBase)
-
-    # Mapping checked by ``SerializerBase._convert_to_dict`` to
-    # decide how to serialise each constructor argument.  ``None``
-    # tells the serialiser to skip that attribute entirely,
-    # a callable value would be invoked to produce the serialised form.
-    # ``interface`` should never be persistent and it will soon be removed.
-    _REDIRECT = {'interface': None}
+    _DEFAULT_PROTECTED_TYPES = (ModelBase,)
 
     def __init__(
         self,
@@ -52,10 +37,9 @@ class CollectionBase(EasyList):
         protected_types: type | Iterable[type] | None = None,
         unique_name: Optional[str] = None,
         display_name: Optional[str] = None,
-        interface: Any = None,  # legacy, should be None and will soon be removed
         data: Optional[Iterable[Any]] = None,
     ):
-        """Create a new collection of EasyScience objects.
+        """Create a new collection of model objects.
 
         Items can be supplied as positional arguments or via the *data*
         iterable.  Both sources are merged in order.
@@ -63,16 +47,14 @@ class CollectionBase(EasyList):
         If the first positional argument is a string and *name* is not given,
         it is consumed as the collection name.
 
-        :param items: EasyScience objects to include in the collection.
+        :param items: Model objects to include in the collection.
         :param name: Human-readable name for the collection.
         :param protected_types: Allowed item types.  Defaults to
             ``_DEFAULT_PROTECTED_TYPES``.
         :param unique_name: Machine-friendly unique identifier.
         :param display_name: Display label; defaults to *name*.
-        :param interface: Reserved internal attribute — must be ``None``.
         :param data: Additional iterable of items to append.
-        :raises AttributeError: If *interface* is not ``None`` or an item
-            fails type validation.
+        :raises AttributeError: If an item fails type validation.
         """
         if items and isinstance(items[0], str) and name is None:
             name = items[0]
@@ -83,15 +65,9 @@ class CollectionBase(EasyList):
 
         super().__init__(unique_name=unique_name, display_name=display_name)
 
-        if interface is not None:
-            raise AttributeError(
-                'Given kwarg: `interface`, is an internal attribute. Please rename.'
-            )
-
         self._protected_types = self._normalize_protected_types(protected_types)
         self._name = name if name is not None else self.display_name
-        self.user_data: dict[str, Any] = {}
-        self.interface = None
+        self._data: list[Any] = []
 
         all_items = self._collect_items(items, data=data)
         for item in all_items:
@@ -99,12 +75,9 @@ class CollectionBase(EasyList):
                 self._validate_item(item)
             except TypeError as exc:
                 raise AttributeError(
-                    'A collection can only be formed from easyscience objects.'
+                    'A collection can only be formed from model objects.'
                 ) from exc
             if item in self:
-                warnings.warn(
-                    f'Item with unique name "{self._get_key(item)}" already in CollectionBase, it will be ignored'
-                )
                 continue
             self._data.append(item)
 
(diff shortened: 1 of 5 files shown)


rozyczko:
Another question:
Currently, the code allows backward compatibility, so old Parameter and Descriptor objects can be used alongside the new ones based on `NewBase`.
Does it make sense to keep this compatibility layer for the interim process or should we completely forget about `ObjBase` derived classes?

rozyczko:
> Another question: Currently, the code allows backward compatibility, so old Parameter and Descriptor objects can be used alongside the new ones based on `NewBase`. Does it make sense to keep this compatibility layer for the interim process or should we completely forget about `ObjBase` derived classes?

I don't think it makes sense to carry over functionality which will be removed in the next release.
Removed all vestiges of BasedBase dependencies, inclusing methods for backward compatibility with the old ModelCollection.

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+            name = items[0]
+            items = items[1:]
+
+        if display_name is None and name is not None:
+            display_name = name
+
+        super().__init__(unique_name=unique_name, display_name=display_name)
+
+        self._protected_types = self._normalize_protected_types(protected_types)
+        self._name = name if name is not None else self.display_name
```
henrikjacobsenfys, line 70: If `display_name` and `name` are both None, then `self._name` is also None. Is that desired, or should there be a fallback?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+        """Human-readable name of the collection."""
+        return self._name
+
+    @name.setter
+    def name(self, new_name: str) -> None:
+        """Set the collection name and sync *display_name*."""
+        if not isinstance(new_name, str):
+            raise TypeError('Name must be a string')
+        self._name = new_name
+        self.display_name = new_name
```
henrikjacobsenfys, line 96: Is that actually desired behavior? I would think that people want to keep `name` and `display_name` separate?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+
+    def __getitem__(self, idx: int | slice | str) -> Any:
+        """Retrieve items by integer index, slice, unique-name key, or
+        display name.
+
+        String lookup first tries the unique_name, then falls back to matching
+        on the item's *name* attribute.  If multiple items share the same name
+        a new ``CollectionBase`` containing all matches is returned.
+        """
+        if isinstance(idx, bool):
```
henrikjacobsenfys, line 108: I think we need to raise an error if idx is not int, slice or str

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+
+        When *value* is a plain number and the existing item has a ``value``
+        attribute, the number is assigned to ``item.value`` (in-place update).
+        Otherwise the item is replaced after type validation.
+        """
+        if isinstance(idx, int) and isinstance(value, Number):
+            item = self[idx]
+            if not hasattr(item, 'value'):
+                raise NotImplementedError(
+                    'At the moment only numerical values or model objects can be set.'
```
henrikjacobsenfys: Is this the right error? Isn't the error here that the user is trying to set the `value` of an object that doesn't have a `value` property?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+                )
+            item.value = value
+            return
+        if isinstance(idx, int):
+            self._validate_item(value)
+            if value is not self._data[idx] and value in self:
```
henrikjacobsenfys, line 141: Not sure I understand what happens - is this right?
We have a collection with e.g. two items in it, `item0` and `item1`. User writes
`mycollection[0]=item1`
and nothing happens, since `item1` is already in the collection, but at a different index. 
If this is the case then maybe we want to warn the user that their assignment didn't do anything?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+                return
+            self._data[idx] = value
+            return
+        if isinstance(idx, slice):
+            if not isinstance(value, Iterable):
+                raise TypeError('Value must be an iterable for slice assignment')
+            new_values = list(value)
+            for v in new_values:
+                self._validate_item(v)
+            self._data[idx] = new_values
```
henrikjacobsenfys, line 161: What happens here if `new_values` is a list of the same thing? Using the example above, if `new_values=[item0,item0]`?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+    def __delitem__(self, idx: int | slice | str) -> None:
+        """Delete an item by index, slice, or unique_name."""
+        if isinstance(idx, (int, slice)):
+            del self._data[idx]
+        elif isinstance(idx, str):
+            for i, item in enumerate(self._data):
+                if self._get_key(item) == idx:
+                    del self._data[i]
+                    return
+            raise KeyError(f'No item with unique name "{idx}" found')
```
henrikjacobsenfys: Should we also allow deleting by name as a fallback like in `__getitem__`?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+
+    def insert(self, index: int, value: Any) -> None:
+        """Insert *value* before *index*, validating it against
+        protected types.
+        """
+        try:
+            self._validate_item(value)
+        except TypeError as exc:
+            raise AttributeError('Only model objects can be put into a CollectionBase') from exc
+        if value in self:
```
henrikjacobsenfys, line 196: Like before, this should probably inform the user that nothing happened and why

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+            raise TypeError('Index must be an int, slice, or str')
+
+    def __len__(self) -> int:
+        """Return the number of items in the collection."""
+        return len(self._data)
+
+    def insert(self, index: int, value: Any) -> None:
+        """Insert *value* before *index*, validating it against
+        protected types.
+        """
```
henrikjacobsenfys, line 189: Do we need a check of the type of `index`?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+    def sort(self, key=None, reverse: bool = False) -> None:
+        """Sort items in place.
+
+        :param key: Single-argument function used to extract a comparison key.
+        :param reverse: If ``True``, sort in descending order.
+        """
+        self._data.sort(key=key, reverse=reverse)
+
+    def index(self, value: Any, start: int = 0, stop: int | None = None) -> int:
+        if stop is None:
```
henrikjacobsenfys, line 225: Could use a docstring

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+        if isinstance(value, str):
+            for i in range(start, min(stop, len(self._data))):
+                if self._get_key(self._data[i]) == value:
+                    return i
+            raise ValueError(f'{value} is not in CollectionBase')
+        return self._data.index(value, start, stop)
+
+    def pop(self, index: int | str = -1) -> Any:
+        """Remove and return an item at the given index or
+        unique_name.
```
henrikjacobsenfys: or name?

src/easyscience/base_classes/collection_base_easylist.py
```diff
@@ -0,0 +1,408 @@
...
+                {'@module': cls_.__module__, '@class': cls_.__name__}
+                for cls_ in self._protected_types
+            ]
+        dict_repr['data'] = [item.to_dict(skip=skip) for item in self._data]
+        return dict_repr
+
+    @classmethod
+    def from_dict(cls, obj_dict: dict[str, Any]) -> CollectionBase:
+        """Reconstruct a ``CollectionBase`` from a dict produced by
+        :meth:`to_dict`.
```
henrikjacobsenfys, line 323: :meth: should probably be :math:

henrikjacobsenfys (commented):
This may be a stupid question, but shouldn't this inherit from or use EasyList? There is at least some duplication of logic.

Otherwise a handful of nitpicks and questions to consider.

src/easyscience/base_classes/collection_base_easylist.py
rozyczko: I don't thing this is an issue. When `name` is None, `self._name` falls back to `self.display_name`, which in turn falls back to `self.unique_name` (auto-generated) in `new_base.py`: 83-86. So `_name` is never None.

src/easyscience/base_classes/collection_base_easylist.py
rozyczko: `:meth:` is actually correct here. It's a Sphinx cross-reference role pointing to the `to_dict` method.
`:math:` is for LaTeX mathematical notation, which would be wrong in this context :)

src/easyscience/base_classes/collection_base_easylist.py
rozyczko: This is intentional. The constructor already syncs them (`if display_name is None and name is not None: display_name = name` )
The setter keeps them in sync for consistency. If a user explicitly sets a different [display_name] via the constructor, the two diverge until `name` is set again.

This could be revisited if use cases emerge where they need to stay independent after initial construction.

rozyczko:
> This may be a stupid question, but shouldn't this inherit from or use EasyList? There is at least some duplication of logic.

`CollectionBase` inherits from `ModelBase` (which provides parameter aggregation, get_all_parameters(), get_free_parameters(), etc.), while `EasyList` inherits from `NewBase`.

EasyList is a pure typed container. It holds a flat list of `NewBase` items in `_data`, enforces protected_types, and provides list operations + unique-name lookup. It has no awareness of models, parameters, or fitting — it's essentially a `list[T]`with type safety and serialization.

The new `CollectionBase` is a model-aware collection. Inherits from `ModelBase` and gets the full parameter aggregation API (`get_*_parameters`), running recursively. It also adds `name`, display-name lookups, and can have its own deserialization. This means a `CollectionBase` can be directly used in fitting.

src/easyscience/base_classes/collection_base_easylist.py
rozyczko: Added warning

src/easyscience/base_classes/collection_base_easylist.py
henrikjacobsenfys: I may be immature for finding this funny

henrikjacobsenfys:
Looks good to me now

henrikjacobsenfys (approved):

damskii9992 (commented):
Pretty sure you are making this `CollectionBase` needlessly complicated. It is considerably less code to simply inherit from `EasyList`, overwrite the default protected_type to `ModelBase` rather than `NewBase` and then just implement the `get_all_variables` methods.

It should also be much more maintainable. We might need to make the class also inherit from `ModelBase` just to make sure it is registered as an instance of a `ModelBase` class, but since you overwrite all the methods, it would only be for type-checking.

Also, why does the class even have a `name` attribute? Didn't we want to get rid of that?

rozyczko:
Superseded by #239

PR #229, opened 2026-03-30
Python 3.14 support

rozyczko:
rebased against current develop
h5py with 3.14 support is already on pypi, so this should be straightforward.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -5,7 +5,7 @@ _src_path: gh:easyscience/templates
 lib_docs_url: https://easyscience.github.io/core
 lib_doi: 10.5281/zenodo.18163581
 lib_package_name: easyscience
-lib_python_max: '3.13'
+lib_python_max: '3.14'
 lib_python_min: '3.11'
 lib_repo_name: core
 project_contact_email: support@easyscience.org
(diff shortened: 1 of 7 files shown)


rozyczko:
Still fails!
This time, some obscure dependency of one of the modules fails to build properly on 3.14
`untokenize` fails.
It's known and reported https://github.com/myint/untokenize/issues/4

This dependency is pulled by `docformatter`, used in `dev` environment, so quite pervasive.
`docformatter` is also aware of this issue https://github.com/PyCQA/docformatter/issues/327 which they apparently fixed but not yet released.

rozyczko:
Can we replace `docformatter` with `ruff format`?
just add `docstring-code-format = true`

rozyczko:
From @AndrewSazonov:

> 
> Ruff format is not really a replacement for docformatter - the latter handles some more important docstring-specific formatting checks.
> I’m aware of the current issue with docformatter and its dependency not supporting Python 3.14. Also, docformatter doesn’t handle Args and Returns sections in Google-style docstrings.
> Because of that, I started looking into alternatives. It turns out that Google style is not that well supported in general, while NumPy style has much better support.
> I tested this in the templates (in a separate branch) and found NumPy style to be better covered. So I switched to another package, format-docstring, instead of docformatter. It has good support for NumPy-style docstrings.
> This means that moving to Python 3.14 should be possible by parallel switching the docstring formatter and reformatting docstrings to a better-supported style.

PR #231, opened 2026-04-09
Fixed minimizer behaviour on exceeded iterations

### LINKED ISSUE #220: Inconsistent behaviour when fits reach max evaluations
We currently support 3 different minimization engines, all these 3 engines behaves differently when the fit does not converge:

## DFO
DFO will throw a `FitError` with an appropriate error message:
Image

## LMFit
LMFit does not throw or warns, but its `success` flag is properly set to `False`

## Bumps
Bumps does not throw or warns, AND its `success` flag is wrongly set to `True` because the fit works and doesn't throw anywhere (it JUST reaches max_iterations).

While not a bug, I also wish to have `evaluations` as a result in our `Result` object, so that it can easily be inspected for profiling without having to know its underlying name in our minimizer backends.

rozyczko:
This adds improvements to the fitting result reporting and evaluation tracking across all minimizer engines, ensuring that the number of function evaluations and relevant messages are consistently captured and propagated. 

* All minimizer engines (`Bumps`, `DFO`, `LMFit`) now populate `n_evaluations` (number of function evaluations) and `message` (status or error message) fields in `FitResults`, and these fields are propagated in multi-dataset fitting. 

* The `Bumps` minimizer uses a new `_EvalCounter` wrapper to count function evaluations and detect when the maximum evaluation budget is reached, updating the `success` flag and message accordingly. 

* The `DFO` minimizer now raises `FitError` only for real failures, and considers both success and max function evaluation warnings as successful fits, improving error reporting and testability.


tests/integration/fitting/test_fitter.py
```diff
@@ -207,14 +207,48 @@ def test_basic_max_evaluations(fit_engine):
...
+        try:
+            f.switch_minimizer(fit_engine)
+        except AttributeError:
+            pytest.skip(msg=f'{fit_engine} is not installed')
+    f.max_evaluations = 3
+    result = f.fit(x=x, y=y, weights=weights)
+
+    assert result.success is False
+    assert result.n_evaluations is not None
+    assert result.n_evaluations > 0
```
henrikjacobsenfys, line 249: Would it make sense to check that n_evaluations matches the budget?

henrikjacobsenfys (approved):
Looks good to me apart from a tiny nitpick.

This may also be a good opportunity to make a `__repr__` method for the `FitResults` class?

rozyczko:
> This may also be a good opportunity to make a `__repr__` method for the `FitResults` class?

Good idea!
Added these
```python
   lines = [
        f"FitResults(success={self.success}",
        f"  n_pars={self.n_pars}, n_points={len(self.x)}",
        f"  chi2={self.chi2:.4g}, reduced_chi={self.reduced_chi:.4g}",
        f"  n_evaluations={self.n_evaluations}",
        f"  minimizer={self.minimizer_engine.__name__ if self.minimizer_engine else None}",
    ]
```

src/easyscience/fitting/minimizers/utils.py
```diff
@@ -35,9 +37,44 @@ def __init__(self):
...
+
+        try:
+            n_points = len(self.x)
+        except TypeError:
+            n_points = 0
+
+        lines = [
+            f'FitResults(success={self.success}',
+            f'  n_pars={self.n_pars}, n_points={n_points}',
+            f'  chi2={chi2}, reduced_chi={reduced}',
```
AndrewSazonov: Does `reduced_chi` mean `reduced_chi2`? If yes, would it make sense to rename it to `reduced_chi2` for consistency with `chi2`?

rozyczko: good point. Fixed.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -275,6 +304,12 @@ def _gen_fit_results(self, fit_results, **kwargs) -> FitResults:
         results.y_obs = self._cached_model.y
         results.y_calc = self.evaluate(results.x, minimizer_parameters=results.p)
         results.y_err = self._cached_model.dy
+        results.n_evaluations = n_evaluations
+        results.message = (
+            f'Fit stopped: reached maximum evaluations ({max_evaluations})'
+            if stopped_on_budget
+            else ''
+        )
```
damskii9992, line 312: Just adding my comment from the other PR here, since this is actually where it belongs. In my attempted implementation, I found that the number of iterations is actually present in the BUMPS `fit_results` simply under `iterations`. So maybe you don't need the `_EvalCounter` at all and can just use this?
```python
        # See: https://github.com/bumps/bumps/blob/master/bumps/fitters.py#L625 for the rationale behind this code.
        if max_evaluations is None:
            for bumps_fitter in FITTERS:
                if bumps_fitter.id == self._method:
                    max_evaluations = bumps_fitter.settings[0][1] # This is the default number of steps for the method
                    break
        if tolerance is None:
            for bumps_fitter in FITTERS:
                if bumps_fitter.id == self._method:
                    tolerance = min(bumps_fitter.settings[1][1], bumps_fitter.settings[2][1]) # These are ftol and xtol
                    break
        if results.iterations == max_evaluations-1:
            warnings.warn(f'Fit did not reach the desired tolerance of {tolerance} within the maximum number of evaluations of {max_evaluations}. '  # noqa: E501
                          f'Consider increasing the maximum number of evaluations or adjusting the tolerance.')
            results.success = False
```

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -220,11 +224,14 @@ def _gen_fit_results(self, fit_results, weights, **kwargs) -> FitResults:
         results.y_obs = self._cached_model.y
         results.y_calc = self.evaluate(results.x, minimizer_parameters=results.p)
         results.y_err = weights
+        results.n_evaluations = int(fit_results.nf)
```
damskii9992, line 227: shouldn't this be `fit_results.nfev`?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -208,7 +212,7 @@ def _gen_fit_results(self, fit_results, weights, **kwargs) -> FitResults:
         for name, value in kwargs.items():
             if getattr(results, name, False):
                 setattr(results, name, value)
-        results.success = not bool(fit_results.flag)
+        results.success = fit_results.flag == fit_results.EXIT_SUCCESS
```
damskii9992, line 215: Can you explain this? What is `fit_results.EXIT_SUCCESS`?

damskii9992: And should we raise a warning if the fit did not succeed?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -258,10 +265,10 @@ def _dfo_fit(
 
         results = dfols.solve(model, pars_values, bounds=bounds, **kwargs)
 
-        if 'Success' not in results.msg:
-            raise FitError(f'Fit failed with message: {results.msg}')
+        if results.flag in {results.EXIT_SUCCESS, results.EXIT_MAXFUN_WARNING}:
```
damskii9992, line 268: Could you also explain this? Maybe add a comment?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -122,6 +122,10 @@ def fit(
             model_results = self._dfo_fit(self._cached_pars, model, **kwargs)
             self._set_parameter_fit_result(model_results, stack_status)
             results = self._gen_fit_results(model_results, weights)
+        except FitError:
+            for key in self._cached_pars.keys():
+                self._cached_pars[key].value = self._cached_pars_vals[key][0]
+            raise
```
damskii9992, line 128: Codecov says this misses a test :)

src/easyscience/fitting/minimizers/minimizer_lmfit.py
```diff
@@ -298,6 +298,8 @@ def _gen_fit_results(self, fit_results: ModelResult, **kwargs) -> FitResults:
         # results.goodness_of_fit = fit_results.chisqr
         results.y_calc = fit_results.best_fit
         results.y_err = 1 / fit_results.weights
+        results.n_evaluations = fit_results.nfev
+        results.message = fit_results.message
```
damskii9992, line 302: We should probably either raise an error or a warning when `max_evaluations` have been reached.

tests/integration/fitting/test_fitter.py
damskii9992, line 249: Since the fit reached max evaluations, shouldn't this be:
```python
assert result.n_evaluations == 3
```

tests/unit/fitting/minimizers/test_minimizer_bumps.py
```diff
@@ -193,28 +197,68 @@ def test_gen_fit_results(self, minimizer: Bumps, monkeypatch):
...
-        assert domain_fit_results.success == True
+        assert domain_fit_results.success == False
         assert domain_fit_results.y_obs == 'y'
         assert domain_fit_results.x == 'x'
         assert domain_fit_results.p == {'ppar_1': 'par_value_1', 'ppar_2': 'par_value_2'}
         assert domain_fit_results.p0 == 'p_0'
         assert domain_fit_results.y_calc == 'evaluate'
         assert domain_fit_results.y_err == 'dy'
+        assert domain_fit_results.n_evaluations == 7
+        assert domain_fit_results.message == 'Fit stopped: reached maximum evaluations (3)'
```
damskii9992, line 221: Huh? max evaluations was 3 but it got 7 iterations? What?

damskii9992 (commented):
Generally looks good. I have some questions/potential changes before approval.
Also, do we want to raise an error or a warning when a fit does not converge withing max_iterations?

rozyczko:
> Generally looks good. I have some questions/potential changes before approval. Also, do we want to raise an error or a warning when a fit does not converge withing max_iterations?

I would prefer to just emit the warning and let the caller check the results object for convergence.
Or maybe we should all have a discussion about that?

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: question addressed in the other PR

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Added

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: DFO's version of `nfev` is `nf`

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: In DFO-LS, `EXIT_SUCCESS` is 0 when “objective is sufficiently small” or “rho has reached rhoend”. Positive codes are warning exits, and negative codes are hard errors.

whether to raise a warning: not for every non-success. Hard failures already become an exception in `minimizer_dfo`, so warning on those would be redundant. But for warning-style exits that still return results, especially EXIT_MAXFUN_WARNING, adding a UserWarning would be reasonable if we want consistency with BUMPS, which warns when it hits the evaluation budget.

rozyczko: Added warning and unit test

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: Added comment

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: test added. you know we shouldn't really aim for 100% coverage? :)

src/easyscience/fitting/minimizers/minimizer_lmfit.py
rozyczko: As mentioned earlier, I would prefer to let the caller decide on what to do with this information. Let's spit a warning and be gracious wrt. max_evaluations.

tests/unit/fitting/minimizers/test_minimizer_bumps.py
rozyczko: ```minimizer._eval_counter = MagicMock(count=7)``` in line 205.

tests/integration/fitting/test_fitter.py
rozyczko: `n_evaluations` is the actual backend-reported evaluation count, not necessarily the configured budget. In this case, `n_evaluations` is 10.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -28,6 +31,19 @@
...
+    def __init__(self, fn: Callable):
+        self._fn = fn
+        self.count = 0
+        self.__name__ = getattr(fn, '__name__', self.__class__.__name__)
+        self.__signature__ = inspect.signature(fn)
+        functools.update_wrapper(self, fn)
+
+    def __call__(self, *args, **kwargs):
+        self.count += 1
+        return self._fn(*args, **kwargs)
```
damskii9992, line 44: Thinking about this now, but you're tracking fit iterations by tracking how many times the model function is evaluated. This is not a correct representation of the number of fit iterations, as some fitting routines evaluate the model function multiple times per fitting step in order to best choose the next fitting step.
Optimizers like BUMPS, for example, run multiple model evaluations at the start of each fit.

This might be the best we CAN do, if we can't get the actual fit iterations from the back engine, but it's worth mentioning.

damskii9992: You might want to add to the integration tests where we actually run the fits and check the iteration counters.

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: Okay, but what about the `fit_results.flag`? I assume this can be other values than 0. Do we risk this returning `True` when the flag and EXIT_SUCCESS are identical values different from 0? What is `EXIT_SUCCESS` value when we hit a `EXIT_MAXFUN_WARNING`?

rozyczko: No, there is no risk from the value being nonzero by itself.
This is a straight equality check:
```
results.success = fit_results.flag == fit_results.EXIT_SUCCESS
```
It only returns True when the flag exactly matches success const. It does not depend on success being 0 specifically. Even if a library used 137 for success, this comparison would still be correct as long as fit_results.flag and fit_results.EXIT_SUCCESS were both 137 for a real success.

For DFO we have:
```
EXIT_SUCCESS = 0
EXIT_MAXFUN_WARNING = 1
EXIT_SLOW_WARNING = 2
EXIT_FALSE_SUCCESS_WARNING = 3
```
hard errors have negative values

The line does what we want:
normal success -> True
maxfun warning -> False
any other non-success flag -> False

I added a comment above the assignment to spell it out.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: This is the best we can currently do.

The actual iteration counters might differ from machine to machine, from OS to OS, I don't think it's going to be easy to catch the right situation, but I'll give it a try

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: But what about `fit_results.flag`, what values can that take? Only 0 and 1?

tests/integration/fitting/test_fitter.py
damskii9992: But doesn't that mean that setting `max_evaluations` doesn't currently work? If the max is 3, the number of iterations should be 3 unless it converged in under 3 iterations (doubtful). I guess this relates to my earlier comment that BUMPS runs multiple function evaluations before actually starting the fit.

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: > But what about `fit_results.flag`, what values can that take? Only 0 and 1?

Line 271: DFO results  object is returned, if results.flag is either 0 (success) or 1 (MAXFUN_WARNING), the results object is returned. Otherwise it is assumed the fit failed and assert raised.

So in line 219, where this is actually being checked, flag_results.flag is only 0 or 1.

tests/integration/fitting/test_fitter.py
rozyczko: yes, BUMPS is interesting.

Our `n_evaluations` comes from _EvalCounter.call, which counts every **objective** invocation.
Simplex nethod evaluates multiple points per optimizer iteration + does work before the main loop starts by building the initial simplex.

So with steps = 3 getting 10 `n_evaluations` is plausible. The solver is not running 10 iterations but rather using up to 3 simplex steps, and those steps require multiple objective calls.

`f.max_evaluations = 3` for BUMPS currently means: maximum optimizer steps
`result.n_evaluations`  means: actual objective-function calls observed.
That is why I changed the BUMPS message to say “maximum optimizer steps” rather than “maximum evaluations”.

The real issue is naming: in BUMPS, `max_evaluations` is currently mapped to steps, so the public API name is more general than the actual behavior.

And we can't set the actual mac number of optimizer steps explicitly...

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: I just checked, the number of iterations used by Bumps is saved in the `engine_result.nit`:

Image

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: Ahh, now I think I get it. DFO sets `fit_results.flag` to some integer depending on if the fit failed (negative), succeeded cleanly (0) or succeeded with a warning (positive). To remember what each of these integers means, they're also assigned to the attributes `EXIT_SUCCESS`, `EXIT_MAXFUN_WARNING` etc.
In other words, the flag attribute is a variable and the `EXIT` attributes are static constants.
So when you compare, you're just checking the flag against the static attribute.
I guess this is done so that they can change the flag integers for different fit terminations without breaking code, or for more easily readable syntax . . . Hmm, interesting.
Could you boil my realization down and add as a comment? Either instead of your comment or as an addendum :)

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: I guess we may use `nit`, assuming BUMPS is consistent with it.

rozyczko:
I assume that this is now ready for merge

PR #232, opened 2026-04-13
Interim updates

rozyczko:
This pull request adds support for progress callbacks during fitting operations. The main goal is to allow users to receive progress updates and optionally cancel fits in progress. Added `progress_callback` parameter and a way to handle cancellation gracefully. 

* Added an optional `progress_callback` parameter to the fit methods in `Fitter`, `MinimizerBase`, `MultiFitter` and all minimizer subclasses, allowing users to receive iterative progress updates and cancel fits. 

* Implemented progress callback integration in `LMFit`: added `_create_iter_callback` to wrap the user callback, constructed detailed progress payloads, and enabled fit cancellation by raising a new `FitCancelled` exception. Ensured parameter values are restored on cancellation or error.

* Added the `FitCancelled` exception class to signal user-requested cancellation, and ensured proper restoration of parameter values on cancellation or error.


rozyczko:
Required for https://github.com/easyscience/EasyReflectometryApp/issues/290

src/easyscience/fitting/minimizers/minimizer_base.py
```diff
@@ -58,6 +58,10 @@ def enum(self) -> AvailableMinimizers:
     def name(self) -> str:
         return self._minimizer_enum.name
 
+    def _restore_parameter_values(self) -> None:
+        for key in self._cached_pars.keys():
+            self._cached_pars[key].value = self._cached_pars_vals[key][0]
```
damskii9992, line 63: You probably also want to restore the error/variance here.

src/easyscience/fitting/minimizers/minimizer_base.py
```diff
@@ -69,6 +73,7 @@ def fit(
         method: Optional[str] = None,
         tolerance: Optional[float] = None,
         max_evaluations: Optional[int] = None,
+        progress_callback: Optional[Callable[[dict], Optional[bool]]] = None,
```
damskii9992, line 76: We don't type hint optional parameters with `Optional` anymore, use `| None`.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -28,6 +32,43 @@
...
+    """
+
+    def __init__(self):
+        self.last_step = 0
+
+    def config_history(self, history):
+        history.requires(step=1)
+
+    def __call__(self, history):
+        self.last_step = int(history.step[0])
```
damskii9992, line 46: Why not just use the actual class implemented in Bumps? `StepMonitor` in `fitters` seems to do exactly what you want.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -28,6 +32,43 @@
...
+
+    def config_history(self, history):
+        history.requires(step=1, point=1, value=1)
+
+    def __call__(self, history):
+        payload = self._owner._build_progress_payload(
+            problem=self._problem,
+            iteration=int(history.step[0]),
+            point=np.asarray(history.point[0]),
+            nllf=float(history.value[0]),
```
damskii9992, line 63: nllf? What is that supposed to mean?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -124,10 +163,8 @@ def fit(
         minimizer_kwargs.update(engine_kwargs)
 
         if tolerance is not None:
-            minimizer_kwargs['ftol'] = tolerance  # tolerance for change in function value
-            minimizer_kwargs['xtol'] = (
-                tolerance  # tolerance for change in parameter value, could be an independent value
-            )
+            minimizer_kwargs['ftol'] = tolerance
+            minimizer_kwargs['xtol'] = tolerance
```
damskii9992, line 163: You probably shouldn't remove the comments here. They're good to have to remember what the difference between xtol and ftol is without having to find it in the Bumps source code. ( I really hate that they don't have a documentation).

src/easyscience/fitting/minimizers/minimizer_base.py
damskii9992: And this method should be reused in the `fit` method of all the minimizers when fitting fails.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
         self._p_0 = {f'p{key}': self._cached_pars[key].value for key in self._cached_pars.keys()}
 
         problem = FitProblem(model)
+
+        method_str = method_dict.get('method', self._method)
+        fitclass = self._resolve_fitclass(method_str)
+
+        step_counter = _StepCounterMonitor()
```
damskii9992, line 179: Should we always monitor the fit? Doesn't this slow down the fit? This should probably be optional so that it can be skipped when run in the library.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -28,6 +32,43 @@
...
+    def __call__(self, history):
+        self.last_step = int(history.step[0])
+
+
+class _BumpsProgressMonitor(Monitor):
+    def __init__(self, owner, problem, callback):
+        self._owner = owner
+        self._problem = problem
+        self._callback = callback
+        self.cancel_requested = False
```
damskii9992: You should probably have a check that the callback provided makes sense and works.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+        fitclass = self._resolve_fitclass(method_str)
+
+        step_counter = _StepCounterMonitor()
+        monitors = [step_counter]
+        progress_monitor = None
+        if progress_callback is not None:
+            progress_monitor = _BumpsProgressMonitor(self, problem, progress_callback)
+            monitors.append(progress_monitor)
+
+        abort_test = (lambda: progress_monitor.cancel_requested) if progress_monitor else None
```
damskii9992: why the new variable name here? It's called `cancel_requested` in the monitor, that is a fitting name, might as well call it the same here.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+
+        mapper = SerialMapper.start_mapper(problem, [])
```
damskii9992, line 184: What does the mapper do?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+        driver = FitDriver(
+            fitclass=fitclass,
+            problem=problem,
+            monitors=monitors,
+            abort_test=abort_test,
+            mapper=mapper,
+            **minimizer_kwargs,
+        )
```
damskii9992, line 191: Ugh, I hate the BUMPS API . . . Now I see why you renamed the variable `abort_test`

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
-            model_results = bumps_fit(problem, **method_dict, **minimizer_kwargs, **kwargs)
-            self._set_parameter_fit_result(model_results, stack_status, problem._parameters)
-            results = self._gen_fit_results(model_results)
+            x_result, fx = driver.fit()
+            if progress_monitor is not None and progress_monitor.cancel_requested:
+                raise FitCancelled()
+            self._set_parameter_fit_result(x_result, driver, stack_status, problem._parameters)
+            results = self._gen_fit_results(
+                x_result, fx, driver, step_counter.last_step, max_evaluations
+            )
```
damskii9992, line 205: Have you checked that the generated results looks the same now as they did with the simplied fit interface? Because the simplified interface was supposed to make a `Scipy` `OptimizeResults` object with the results, but the `FitDriver` doesn't, as far as I am aware.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+        except FitCancelled:
+            self._restore_parameter_values()
+            raise
         except Exception as e:
-            for key in self._cached_pars.keys():
-                self._cached_pars[key].value = self._cached_pars_vals[key][0]
+            self._restore_parameter_values()
             raise FitError(e)
+        finally:
+            global_object.stack.enabled = stack_status
```
damskii9992, line 210: Why add this now? Does this fix something that was broken before? Wasn't the stack already broken?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -28,6 +32,43 @@
...
+        self._owner = owner
+        self._problem = problem
+        self._callback = callback
+        self.cancel_requested = False
+
+    def config_history(self, history):
+        history.requires(step=1, point=1, value=1)
+
+    def __call__(self, history):
+        payload = self._owner._build_progress_payload(
```
damskii9992, line 59: Children classes are never supposed to know methods of their parent classes. Only parent classes are supposed to know methods 1-layer down of their children. This stinks of bad software architecture and should probably be reworked.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+        raise FitError(f'Unknown BUMPS fitting method: {method}')
+
+    def _build_progress_payload(
+        self, problem, iteration: int, point: np.ndarray, nllf: float
+    ) -> dict:
+        # Use the nllf already computed by the fitter to avoid a costly
+        # model re-evaluation.  For Gaussian likelihoods:
+        #   nllf = sum(residuals**2) / 2  =>  chi2 = 2 * nllf
+        chi2 = 2.0 * nllf
+        dof = problem.dof
```
damskii9992, line 227: What is dof? :)

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+        parameter_values = self._current_parameter_snapshot(problem, point)
+
+        return {
+            'iteration': iteration,
+            'chi2': chi2,
+            'reduced_chi2': reduced_chi2,
+            'parameter_values': parameter_values,
+            'refresh_plots': False,
+            'finished': False,
+        }
```
damskii9992, line 239: What exactly are you trying to accomplish? Is the callback supposed to abort a fit when a certain parameter has reached a certain value?
Also, isn't the reduced chi2 calculated by the fitter? Shouldn't you just use the value they calculated?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -139,22 +176,91 @@ def fit(
...
+            self._restore_parameter_values()
+            raise
         except Exception as e:
-            for key in self._cached_pars.keys():
-                self._cached_pars[key].value = self._cached_pars_vals[key][0]
+            self._restore_parameter_values()
             raise FitError(e)
+        finally:
+            global_object.stack.enabled = stack_status
         return results
```
damskii9992, line 211: Just adding my code here from my attempt at providing a proper error when the fit doesn't succeed. I had this block added just before `return results`.
```python
        # See: https://github.com/bumps/bumps/blob/master/bumps/fitters.py#L625 for the rationale behind this code.
        if max_evaluations is None:
            for bumps_fitter in FITTERS:
                if bumps_fitter.id == self._method:
                    max_evaluations = bumps_fitter.settings[0][1] # This is the default number of steps for the method
                    break
        if tolerance is None:
            for bumps_fitter in FITTERS:
                if bumps_fitter.id == self._method:
                    tolerance = min(bumps_fitter.settings[1][1], bumps_fitter.settings[2][1]) # These are ftol and xtol
                    break
        if results.iterations == max_evaluations-1:
            warnings.warn(f'Fit did not reach the desired tolerance of {tolerance} within the maximum number of evaluations of {max_evaluations}. '  # noqa: E501
                          f'Consider increasing the maximum number of evaluations or adjusting the tolerance.')
            results.success = False
```

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -261,12 +385,19 @@ def _gen_fit_results(self, fit_results, **kwargs) -> FitResults:
...
-        results.success = fit_results.success
+        results.n_evaluations = n_evaluations
+        # Bumps step counter is 0-indexed, so the last step of a budget of N
+        # is N-1.  We therefore compare with ``max_evaluations - 1``.
+        if max_evaluations is not None and n_evaluations >= max_evaluations - 1:
+            results.success = False
+            results.message = f'Maximum number of evaluations ({max_evaluations}) reached'
+        else:
+            results.success = True
+            results.message = 'Optimization terminated successfully'
```
damskii9992, line 382: `max_evaluations` need to be set to the default value if the user didn't provide it. See my other code-block, as we otherwise have the same problem if the fit doesn't converge with the default `max_iterations` :)

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -261,12 +385,19 @@ def _gen_fit_results(self, fit_results, **kwargs) -> FitResults:
         for name, value in kwargs.items():
             if getattr(results, name, False):
                 setattr(results, name, value)
-        results.success = fit_results.success
+        results.n_evaluations = n_evaluations
```
damskii9992, line 374: Just adding this here, but with the previous `fit_results` object, this line should have been:
```python
results.n_evaluations= fit_results.nit
```
But I realize that that isn't possible now that you're using the `FitDriver`

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -64,6 +80,10 @@ def fit(
         method: str = None,
         tolerance: Optional[float] = None,
         max_evaluations: Optional[int] = None,
+        progress_callback: Optional[Callable[[dict], Optional[bool]]] = None,
+        callback: Optional[Callable[[DFOCallbackState], None]] = None,
+        callback_every: int = 1,
+        callback_on_improvement_only: bool = False,
```
damskii9992, line 86: You didn't have these options in BUMPS. I assume these mean that you can choose to only check your callback every `callback_every` iteration (maybe change the name of that attribute)? Just like you can run the callback only in improvement iterations?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -101,9 +121,30 @@ def fit(
         if (weights <= 0).any():
             raise ValueError('Weights must be strictly positive and non-zero.')
 
+        if callback_every < 1:
+            raise ValueError('callback_every must be a positive integer.')
+
+        # Bridge progress_callback into the DFO callback mechanism
+        if progress_callback is not None and callback is None:
+            dof = max(len(x) - len(self._cached_pars), 1)
```
damskii9992, line 129: dof again. What does it mean?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -122,10 +163,14 @@ def fit(
             model_results = self._dfo_fit(self._cached_pars, model, **kwargs)
             self._set_parameter_fit_result(model_results, stack_status)
             results = self._gen_fit_results(model_results, weights)
+        except FitError:
+            self._restore_parameter_values()
+            raise
```
damskii9992, line 168: Is this an empty raise statement? What happens then? O.o

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -101,9 +121,30 @@ def fit(
         if (weights <= 0).any():
             raise ValueError('Weights must be strictly positive and non-zero.')
 
+        if callback_every < 1:
+            raise ValueError('callback_every must be a positive integer.')
```
damskii9992, line 125: what about floats like 1.3?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -220,11 +371,14 @@ def _gen_fit_results(self, fit_results, weights, **kwargs) -> FitResults:
         results.y_obs = self._cached_model.y
         results.y_calc = self.evaluate(results.x, minimizer_parameters=results.p)
         results.y_err = weights
+        results.n_evaluations = int(fit_results.nf)
```
damskii9992, line 374: shouldn't this be `fit_results.nfev`?

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -208,7 +359,7 @@ def _gen_fit_results(self, fit_results, weights, **kwargs) -> FitResults:
         for name, value in kwargs.items():
             if getattr(results, name, False):
                 setattr(results, name, value)
-        results.success = not bool(fit_results.flag)
+        results.success = fit_results.flag == fit_results.EXIT_SUCCESS
```
damskii9992, line 362: Consider raising a warning if the fit did not succeed.

src/easyscience/fitting/minimizers/utils.py
```diff
@@ -35,9 +37,44 @@ def __init__(self):
...
         self.engine_result = None
         self.total_results = None
 
+    def __repr__(self) -> str:
+        engine_name = self.minimizer_engine.__name__ if self.minimizer_engine else None
+        try:
+            chi2_val = self.chi2
+            reduced_val = self.reduced_chi2
+            if not np.isfinite(chi2_val) or not np.isfinite(reduced_val):
+                raise ValueError
```
damskii9992, line 51: This error should probably have some context

damskii9992 (commented):
This PR needs an ADR explaining the software architecture design implemented and the rationale for why this design was chosen etc.
We probably also need a tutorial showcasing how the callback might be used. Both for ourselves as a kind of documentation but also for potential users.
I have not bothered looking at the unit tests.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Bumps `StepMonitor` is designed for logging fit progress to a file (it requires a file handle fid and saves fields like 'step', 'time', 'value', 'point' to disk). It's not intended for in-memory tracking of the last step count.

My `_StepCounterMonitor` is just a lightweight monitor that simply records the most recent step number in self.last_step for the number of evaluations completed. It has no file I/O overhead and serves the specific need of the fitting result generation

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Added check

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Err....
In standard OO design
- Child classes do know and inherit methods from their parent classes. That's the point of inheritance.
- Parent classes generally do not know methods specific to their children (unless overridden, and even then, they shouldn't rely on child-specific behavior to maintain the Liskov Substitution Principle).

rozyczko: If your concern is coupling, maybe a better refactor is to remove knowledge of the owner object, not to force a “parent knows child” rule. This could be redone so `_BumpsProgressMonitor` receives a payload_builder callable instead of owner, then Bumps.fit passes self._build_progress_payload into it.
1. _BumpsProgressMonitor takes problem, callback, and payload_builder.
2. Its call uses `self._payload_builder` to build the payload.
3. Bumps.fit constructs it with `self._build_progress_payload`

Does this look sensible?

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: using bumps terminology so it is consistent for arguments.
`nllf` = negative log likelihood function.

`chi2 = 2*nllf` essentially

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: well spotted

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Not all monitoring here is the same.

Here, we always attach only the lightweight `_StepCounterMonitor`. It only requests the step field and stores int(history.step[0]). It is then used to report the number of evaluations in the fit results. The heavier progress work is already conditional: `_BumpsProgressMonitor` is only added when `progress_callback` is not None.

rozyczko: We don't always monitor full fit progress; we always attach a minimal step counter because the wrapper currently needs iteration count for FitResults. That should have negligible overhead relative to actual model evaluations.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: removed cancel related functionality

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: cancel moved out

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: `mapper` is just a function that BUMPS uses to evaluate a batch of candidate parameter vectors. I was going to use it for more xomplex things but didn't actually do it.
Removed.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: The proposed public API is `FitResults`, and FitDriver in engine_result is fine.

Have we verified that the public `FitResults` fields produced here match the previous simplified interface for BUMPS? engine_result is backend-specific, so I would only treat that as a regression if callers explicitly depended on a SciPy OptimizeResult.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: yes, if the exception fired, the undo stack was not properly reset.

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: Hmm, we could make a virtual on-memory file to save to, in order to re-use the BUMPS `StepMonitor`, but I'm not sure what is simpler/more maintainable . . .

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: Yeah, parent/child is the wrong terminology here. I meant that a class can know methods of its own attributes (but not its attributes attributes), whereas attributes shoouldn't know about their owners methods and other attributes.
I don't remember what this software design principle is called.

Your proposal looks sensible to fix my concern :)

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: More BUMPS-specific lingo.
DOF - degrees of freedom of the fit. Just a normalization term which BUMPS uses for the fitting problem. dof≈N(data) −N(parameters)

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Yes, BUMPS knows how to compute reduced chi2
No, it is not already returning reduced chi2 in history.value[0]
The wrapper should probably call problem.chisq(nllf=nllf, norm=True) rather than manually doing 2 * nllf / dof
Changed.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Will address it in the other PR and merge with this here.

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: Correct
`callback_every` and `callback_on_improvement_only` do exactly control when the user callback is invoked. That is implemented later in minimizer_dfo.py:229:

- `callback_every` throttles callback emission to every N evaluations.
- `callback_on_improvement_only` filters those emissions to cases where the objective improved.

Note, that these options do not change how often the fit itself is evaluated. DFO still evaluates the model every iteration. These flags only control when your wrapper calls back out to user code

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: yes. Missed that. thx

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: This time, we need to explicitly calculate dof, since DFO doesn't do the proper normalization. Since we're reporting chi2 normalized, we need it here as well.
Kept the name to correspond to the standard elsewhere.

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: To re-raise the original `FitError` without altering its traceback after cleanup.
The naked raise is intentional.

That `except FitError:` block is doing one thing before rethrowing: it restores cached parameter values. Using bare raise preserves the original `FitError` and its traceback exactly as-is. If it were changed to `raise FitError(e)`  there, it would wrap or recreate the exception and lose some traceback clarity.

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: No, this is correct for DFO-LS.
nf = number of objective evaluations
DFO doesn't follow SciPy style OptimizeResults API

src/easyscience/fitting/minimizers/utils.py
rozyczko: This ValueError is not a user-facing error. It is raised only to be caught immediately by the `except Exception` in line 54, after which chi2 and reduced_chi2 are rendered as N/A

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: ouch... virtual files are icky. Let's not, unless we REALLY have a very serious requirement.
I would strongly propose against. Strongly enough to mark this resolved.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: replaced as specfied.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: correct.

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: But shouldn't `callback_every` and `callback_on_improvement_only` also be available in the other minimizers? Anyways, since you're separating this out to another PR. We can discuss it there.

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: Ahh, I didn't know about this. I've always just used the syntax:
```python
except FitError as error:
   code
   raise error
```
IMO that is slightly cleaner, but that is just my preference :)

src/easyscience/fitting/minimizers/utils.py
damskii9992: Even then, in the future we might change some code elsewhere to the catch exception and simply re-raise the error here. Without the error context, such re-raises would be hard to diagnose. It never hurts to have an descriptive error-message, even if it isn't currently being used.

rozyczko: I don't think this is necessary, but I will begrudingly accept the request.

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992, line 424: I probably wouldn't save `n_evaluations` here, but rather `iterations`, since what we want to report isn't how many times the fit has evaluated the model function, but rather how many iterations the fit has run.

damskii9992, line 167: I think we should fill in the `tolerance` and `max_evaluations` fields with the default ones from BUMPS, so that the user knows what they are, since they aren't "None":
```python
        # See: https://github.com/bumps/bumps/blob/master/bumps/fitters.py#L625 for the rationale behind this code.
        if max_evaluations is None:
            for bumps_fitter in FITTERS:
                if bumps_fitter.id == self._method:
                    max_evaluations = bumps_fitter.settings[0][1] # This is the default number of steps for the method
                    break
        if tolerance is None:
            for bumps_fitter in FITTERS:
                if bumps_fitter.id == self._method:
                    tolerance = min(bumps_fitter.settings[1][1], bumps_fitter.settings[2][1]) # These are ftol and xtol
                    break
```

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -44,6 +60,25 @@ def __call__(self, *args, **kwargs):
...
+        history.requires(step=1, point=1, value=1)
+
+    def __call__(self, history):
+        payload = self._payload_builder(
+            problem=self._problem,
+            iteration=int(history.step[0]),
+            point=np.asarray(history.point[0]),
+            nllf=float(history.value[0]),
+        )
+        self._callback(payload)
```
damskii9992: Didn't you plan to split this into a separate PR? The callback monitor?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -31,6 +32,21 @@
...
+    def __init__(self):
+        self.last_step = 0
+
+    def config_history(self, history):
+        history.requires(step=1)
+
+    def __call__(self, history):
+        self.last_step = int(history.step[0])
+
+
```
damskii9992: We are only supposed to have 1 class per python file. I could accept 1 additional small class, but now you're reaching 2-3 small additional classes, so I think they belong in separate files in separate sub-folders.

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: Does this change with our latest changes #231? We don't need the minimal step counter for the `FitResults` now, right? Since we get the `n_evaluations` there.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -156,27 +189,100 @@ def fit(
...
+            from scipy.optimize import OptimizeResult
+
+            model_results = OptimizeResult(
+                x=x,
+                dx=driver.stderr(),
+                fun=fx,
+                success=True,
+                status=0,
+                message='successful termination',
+                nit=driver.monitor_runner.history.step[0],
```
damskii9992: Should this `nit` not be the BUMPS `nit`? Is there a difference in this case?

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: You did not separate this out into a different PR, so I guess we should discuss it here. Shouldn't these options also be available in the other minimizers? :)

src/easyscience/fitting/minimizers/minimizer_dfo.py
```diff
@@ -102,9 +122,32 @@ def fit(
         if (weights <= 0).any():
             raise ValueError('Weights must be strictly positive and non-zero.')
 
+        if not isinstance(callback_every, Integral) or isinstance(callback_every, bool):
```
damskii9992: why `Integral` and not just `int`?

src/easyscience/fitting/minimizers/minimizer_lmfit.py
```diff
@@ -158,11 +160,54 @@ def fit(
...
+        return iter_cb
+
+    def _build_progress_payload(self, params, iteration: int, residuals: np.ndarray) -> dict:
+        residual_array = np.asarray(residuals)
+        chi2 = float(np.square(residual_array).sum())
+        varied_parameter_count = sum(
+            1 for parameter in params.values() if getattr(parameter, 'vary', False)
+        )
+        degrees_of_freedom = residual_array.size - varied_parameter_count
+        reduced_chi2 = chi2 / degrees_of_freedom if degrees_of_freedom > 0 else chi2
```
damskii9992, line 193: Same question as for DFO, does LMFIT not calculate chi2 and can't you just use that instead of calculating it again every iteration?

tests/unit/fitting/minimizers/test_minimizer_bumps.py
```diff
@@ -238,6 +257,386 @@ def test_gen_fit_results(self, minimizer: Bumps, monkeypatch):
...
         )
 
+    @pytest.mark.parametrize(
+        'n_evaluations, max_evaluations, expected_success',
+        [
+            (1, 3, True),  # last step (1) < budget-1 (2) => success
+            (2, 3, False),  # last step (2) == budget-1 (2) => budget consumed => failure
+            (3, 3, False),  # last step (3) > budget-1 (2) => failure
+            (0, 1, False),  # 0 >= 0 => failure (budget of 1, step counter 0-indexed)
+            (5, None, True),  # no budget => always success
```
damskii9992: This is wrong. The minimizers have a default budget which is used when we have it set to `None`. So the max budget can still be reached (we just don't know when it is, see my other comment about this).

damskii9992 (commented):
I would still like to see a tutorial or something which showcases how the progress callback might actually be used, not just a `MagicMock` unittest.

src/easyscience/fitting/minimizers/minimizer_bumps.py
henrikjacobsenfys: To be somewhat pedantic: both number of iterations and number of evaluations can be relevant parameters.

tests/unit/fitting/minimizers/test_minimizer_bumps.py
rozyczko: fixed

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: No strong reason. `Integer` take ints but also things like `np.int64` etc. But I guess using `int` is simpler and actually matches the type hint. Fixed

src/easyscience/fitting/minimizers/minimizer_lmfit.py
rozyczko: Nope. lmfit's `iter_cb` signature is `(params, iter, resid, *args, **kwargs)`. There is **no** pre-computed chi2 passed in. LMFIT's MinimizerResult.chisqr is only available **after** the fit completes, not during iteration. So summing `residuals**2` inside the callback is the correct approach.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: it should. Fixed

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Those are very small, single task classes which apply directly to `Bumps` class functionality. Splitting them and placing in say, `utils.py` would mean we could reuse them elsewhere. This is not the case.
If, at some point, we find these can be instantiated by another class, I'd pull them out.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: No, what I split out was the cancellation functionality. The callback monitor has always been a part of this. This is how we get the updates

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: thanks. applied.

rozyczko: Cross-backend consistency (what I said earlier): `LMFit.nfev` and `DFO-LS.nf` both count objective-function calls, so `n_evaluations` matches.
I would like to be consistent and report just this.
I will document explicitly in the BUMPS docstring that for BUMPS this is objective evaluations, not iterations.

rozyczko:
> I would still like to see a tutorial or something which showcases how the progress callback might actually be used, not just a `MagicMock` unittest.

Added `progress-callback.ipynb`

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: Then I would like to require that we ALSO report `iterations` in all of the minimizers. Because I only see a use-case for `iterations`, not for objective evaluations.

henrikjacobsenfys: Users may be interested in the number of objective evaluations if the objective function is very time-consuming to calculate. A completely hypothetical example could be if the function involves a full or partial reduction workflow for an ESS instrument ;)

henrikjacobsenfys: As hinted at above, I'm not against reporting both

rozyczko: I don't really see the need to confuse users but OK. We can add this attribute.
Done

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: I still think I'd prefer them pulled out into separate files in a seperate folder, say `BUMPS_utils`, and then a file for each class. I like clean file structures.

rozyczko: then we should probably spell it out in an ADR

rozyczko: done

src/easyscience/fitting/minimizers/minimizer_dfo.py
damskii9992: Just repeating the question since I never got an answer:

> But shouldn't callback_every and callback_on_improvement_only also be available in the other minimizers?

docs/docs/tutorials/progress-callback.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "for backend in (AvailableMinimizers.LMFit, AvailableMinimizers.Bumps, AvailableMinimizers.DFO):\n",
+    "    res, hist = fit_with(backend)\n",
+    "    results[backend.name] = res\n",
+    "    histories[backend.name] = hist\n",
+    "    print(\n",
+    "        f'{backend.name:>6s}: callback fired {len(hist):4d} times, '\n",
+    "        f'final reduced chi2 = {res.reduced_chi2:.4f}, '\n",
+    "        f'n_evaluations = {res.n_evaluations}'\n",
+    "    )"
+   ]
```
damskii9992, line 297: I ran this notebook, and DFO returned:
final reduced chi2 = 0.0000
Which is obviously wrong . . .

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: DFO is special here, I played with `callback_every` and `_on_improvement_only` but extending these to the other minimizers is not trivial.
`progress_callback` is already nicely harmonized across all three.
I might just remove the DFO specific callbacks.

damskii9992: Maybe branch it off, create a feature issue and link to the branch so that we might revisit it in the future? I can see the appeal of the feature, but I don't think we want to have certain features only available in certain minimizers, it isn't a great user-experience :/

damskii9992 (approved):
Otherwise, pending answers/resolutions to my last 2 comments. I can finally approve this PR. You should also update the ADR to reflect the latest decisions :)

docs/docs/tutorials/progress-callback.ipynb
rozyczko: In dfo, `_gen_fit_results` was assigning `results.y_err = weights`. But `weights` are the residual weights (1/σ - they get multiplied into residuals by `_make_model`, and `FitResults.chi2` computes ((y_obs - y_calc) / y_err)**2, expecting `y_err` to be σ. LMFit and Bumps both correctly invert weights to σ; DFO did not. With weights ≫ 1, chi² collapsed toward 0; with weights = 1, the bug was masked.

src/easyscience/fitting/minimizers/minimizer_dfo.py
rozyczko: Branched off as `dfo_specific_callbacks`.
Removed the dfo specific callbacks in this branch.
Added a feature request https://github.com/easyscience/core/issues/237

PR #239, opened 2026-04-29
Easy collection

rozyczko:
This is the collection list as described in https://github.com/easyscience/core/pull/226
Rebased against the new develop branch and due to the change of the base class, made a separate PR.

For discussion look in: https://github.com/easyscience/core/pull/226

### DIFF
--- a/src/easyscience/base_classes/easy_collection.py
+++ b/src/easyscience/base_classes/easy_collection.py
@@ -122,7 +122,9 @@ class EasyCollection(EasyList[CollectionItem]):
         raise TypeError('Index must be an int, slice, or str')
 
     def insert(self, index: int, value: CollectionItem) -> None:
-        """Insert an item and register it as a child of the collection."""
+        """Insert an item and register it as a child of the
+        collection.
+        """
         if not isinstance(index, int):
             raise TypeError('Index must be an integer')
         self._validate_item(value)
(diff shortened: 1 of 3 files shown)


src/easyscience/base_classes/easy_collection.py
```diff
@@ -0,0 +1,401 @@
...
+        """Get an item by index, slice, or unique name.
+
+        String lookup returns the single item with the matching ``unique_name``.
+        Duplicate ``unique_name`` values are rejected at insertion time, so a
+        successful lookup always resolves to exactly one item.
+        """
+        if isinstance(idx, int):
+            return self._data[idx]
+        if isinstance(idx, slice):
+            return self._new_like(self._data[idx])
```
henrikjacobsenfys, line 76: If I understand correctly, this creates a copy of the original item. I fear this may lead to unintended behaviour with dependent parameters and/or when referring to a specific parameter.

src/easyscience/base_classes/easy_collection.py
```diff
@@ -0,0 +1,401 @@
...
+        return item in self._data
+
+    def index(self, value: CollectionItem | str, start: int = 0, stop: int | None = None) -> int:
+        """Return the index of an item object or the first item with the
+        given unique name.
+        """
+        if stop is None:
+            stop = len(self._data)
+        if isinstance(value, str):
+            for i in range(start, min(stop, len(self._data))):
```
henrikjacobsenfys: What happens here if `stop` is negative?

Also, what happens if `start` or `stop` is not an int?

henrikjacobsenfys (commented):
I don't understand the global object well enough to comment on that aspect, but this looks good to me, with two exceptions

src/easyscience/base_classes/easy_collection.py
rozyczko: great catch!

PR #242, opened 2026-05-04
Add get_all_variables method to EasyList for collection querying

rozyczko:
Added `get_all_variables` method to EasyList to enable proper querying of collections.

This removes the requirement of having a dedicated Collections-type class, as EasyList is already a collection.


src/easyscience/base_classes/easy_list.py
```diff
@@ -191,6 +200,25 @@ def _get_key(self, obj) -> str:
...
+        For each element that is a `ModelBase` instance, the element's
+        own `get_all_variables()` method is called and the results are
+        collected into a single flat list.
+
+        :return: Flat list of all `DescriptorBase` objects from all
+            `ModelBase` elements.
+        :rtype: List[DescriptorBase]
+        """
+        ModelBase = _get_model_base()
+        all_vars: List[DescriptorBase] = []
```
damskii9992, line 208: Should probably be `DescriptorNumber`

src/easyscience/base_classes/easy_list.py
damskii9992: Since we now have a `get_all_variables` method, we should probably inherit from `ModelBase` rather than `NewBase`, since this is the agreement we have made in the API. Anything inheriting from `ModelBase` has a `get_all_variables` method.

src/easyscience/base_classes/easy_list.py
```diff
@@ -17,9 +17,18 @@
...
 from .new_base import NewBase
 
+
+# Import ModelBase lazily to avoid circular imports at module level
+def _get_model_base():
+    from .model_base import ModelBase
+
+    return ModelBase
+
+
```
damskii9992: Considering my other comment, this little utility method should be unnecessary :)

src/easyscience/base_classes/easy_list.py
```diff
@@ -191,6 +200,25 @@ def _get_key(self, obj) -> str:
...
+        :return: Flat list of all `DescriptorBase` objects from all
+            `ModelBase` elements.
+        :rtype: List[DescriptorBase]
+        """
+        ModelBase = _get_model_base()
+        all_vars: List[DescriptorBase] = []
+        for item in self._data:
+            if isinstance(item, ModelBase):
+                all_vars.extend(item.get_all_variables())
+        return all_vars
```
damskii9992, line 212: By inheriting from `ModelBase` we also get the other "get_all" methods for free since we have implemented `get_all_variables` here :)

src/easyscience/base_classes/easy_list.py
damskii9992, line 34: The default `protected_types` should however remain a `NewBase`.

tests/unit/base_classes/test_easy_list.py
```diff
@@ -524,3 +573,73 @@ def test_from_dict_round_trip(self):
...
+
+    def test_get_all_variables_returns_descriptorbase_instances(self):
+        """All returned items should be instances of DescriptorBase."""
+        m1 = MockModel(unique_name='m1', temperature=10, volume=5.0)
+        m2 = MockModel(unique_name='m2', temperature=99, volume=1.5)
+        el = EasyList(m1, m2, protected_types=ModelBase)
+        for v in el.get_all_variables():
+            from easyscience.variable.descriptor_base import DescriptorBase
+
+            assert isinstance(v, DescriptorBase)
```
damskii9992, line 645: How about a test of an `EasyList` containing another `EasyList` with `NewBase` and `ModelBase` elements?

damskii9992 (commented):
Looks good. Nice and simple and short :) Few comments to further improve it.

src/easyscience/base_classes/easy_list.py
rozyczko: ModelBase uses `DescriptorBase` for its `get_all_variables` method. Why claim they're `DescriptorNumber`?

damskii9992: Oh, I guess they don't have to be. I'm still not sure if we ever use any of the other `Descriptor` types, but I guess its fine :)

PR #244, opened 2026-05-06
Bayesian post-analysis using BUMPS DREAM

### LINKED ISSUE #246: Add sampling method for multiffiter/fitter
For running BUMPS/DREAM sampling

rozyczko:
A new Bayesian MCMC sampling interface. This enables users to perform posterior sampling using the BUMPS DREAM sampler directly from both the `MultiFitter` and BUMPS minimizer classes.

* Added a `sample` method to `MultiFitter` that flattens multi-dataset arrays, resolves parameter aliases, and delegates to the minimizer’s sampling method, supporting both 1D and vectorized multi-dimensional fits.
* Implemented a `sample` method in the BUMPS minimizer (`minimizer_bumps.py`) that builds a BUMPS `FitProblem` and runs the DREAM sampler, returning posterior draws, parameter names, sampler state, and log-probabilities.


rozyczko:
I might want to extend this to multiprocessing, since BUMPS already supports this. 
The implementation should speed things up quite noticeably.
(in a separate branch/PR)

henrikjacobsenfys (approved):
Very nice, found no issues except a few things in the notebook:

It's using `ObjBase`, which we're deprecating
Yo ucould add the hash to the pooch retrieve call

henrikjacobsenfys:
Some of the notes don't get rendered right now
image

rozyczko:
> Some of the notes don't get rendered right now image

hmmm
for me it displays the table nicely

Screenshot 2026-05-29 104430

henrikjacobsenfys:
> > Some of the notes don't get rendered right now image
> 
> hmmm for me it displays the table nicely
> 
> Screenshot 2026-05-29 104430

The tables are fine, it's the markdown that I'm unsure of. But could just be a vs code thing

rozyczko:
Fixed both issues

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "- **You care about parameter correlations** &mdash; the joint posterior (Exercise 6) reveals trade-offs between parameters that a single covariance matrix can miss.\n",
+    "- **You want to propagate uncertainty to predictions** &mdash; with the posterior in hand, you can compute credible bands on any function of the parameters (Exercise 7) without linearised error propagation.\n",
+    "\n",
+    "The trade-off is computational cost: MCMC requires thousands of model evaluations, whereas an MLE fit may converge in dozens. For the simple 4-parameter model used here the difference is negligible, but for expensive models it is worth starting with MLE and only switching to MCMC when you need the richer output.\n",
+    "\n",
+    "In this tutorial we re-use the QENS dataset and the Lorentzian-with-resolution model from the [Fitting QENS](fitting-qens.ipynb) tutorial, but instead of returning a single best-fit value with a symmetric error bar we will draw thousands of samples from the posterior."
```
damskii9992: I would probably have a more thorough documentation to explain what Bayesian analysis is elsewhere, as this is kinda short and hard to understand unless you already know Bayesian analysis, and then in this notebook not bother explaining it, maybe just reference that other documentation. But maybe that is a task for a future easyscience documentation effort?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "filename = fetch_data(\n",
+    "    '4-reduction/energy_transfer_QENS_unknown_quasi_elastic_many_neutrons.dat',\n",
+    "    known_hash='sha256:e49fa9a1d2ef5eeb714524903e8ffa9de6616e5a299a799cbb0b2f3ee8fd459a',\n",
+    ")\n",
+    "\n",
+    "\n",
+    "def load(filename: str):\n",
+    "    \"\"\"Load data from file and filter NaN values.\"\"\"\n",
+    "    x, y, e = np.loadtxt(filename, unpack=True)\n",
+    "    sel = np.isfinite(y)\n",
```
damskii9992: x and y are okay(ish) to be single letter variables. e should be error and sel should be selection.
Death to abbreviations and short variables in code!
Also, what kind of file does this load? .xye files? .txt. files? A user might try this notebook with their own files, so please make it clear what this loader can load.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "source": [
+    "def fetch_data(name: str, known_hash: str) -> str:\n",
+    "    \"\"\"\n",
+    "    Fetch pre-prepared data from a remote source and return the path to the file.\n",
+    "    \"\"\"\n",
+    "    import pooch\n",
+    "\n",
+    "    return pooch.retrieve(\n",
+    "        url=f'https://public.esss.dk/groups/scipp/dmsc-summer-school/2025/{name}',\n",
+    "        known_hash=known_hash,\n",
```
damskii9992, line 87: We should probably have a .datasets folder in `easyscience` to make import of data-files for tutorials easy. A tutorial really shouldn't have cells like these . . .

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "    x, y, e = np.loadtxt(filename, unpack=True)\n",
+    "    sel = np.isfinite(y)\n",
+    "    return x[sel], y[sel], e[sel]\n",
+    "\n",
+    "\n",
+    "omega, intensity_obs, di = load(filename)\n",
+    "\n",
+    "# Restrict to the region of interest, as in the QENS tutorial.\n",
+    "sel = (omega > -0.06) & (omega < 0.06)\n",
+    "omega, intensity_obs, di = omega[sel], intensity_obs[sel], di[sel]"
```
damskii9992: sel? intensity_obs? di?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, ax = plt.subplots()\n",
+    "ax.errorbar(omega, intensity_obs, di, fmt='.')\n",
+    "ax.set(xlabel='$\\\\omega$/meV', ylabel='$I(\\\\omega)$')\n",
+    "plt.show()"
```
damskii9992, line 136: Do we wanna use plopp or plotly instead?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "\n",
+    "$$\n",
+    "I(\\omega) = \\frac{A\\gamma}{\\pi\\big[(\\omega - \\omega_0)^2 + \\gamma^2\\big]} \\;\\ast\\; \\mathcal{N}(0, \\sigma),\n",
+    "$$ (model)\n",
+    "\n",
+    "where $A$ is a scale factor, $\\gamma$ is the Lorentzian half-width at half-maximum, $\\omega_0$ is the centre offset and $\\sigma$ is the width of the Gaussian resolution kernel.\n",
+    "\n",
+    "```{note}\n",
+    "The convolution below uses `np.convolve(..., 'same')` for simplicity. In `'same'` mode, `numpy` pads the signal edges with zeros, which can introduce small artefacts at the boundaries. Because our Gaussian kernel ($\\sigma \\approx 10^{-3}$) is much narrower than the data range ($\\pm 0.06$ meV), these edge effects are negligible here. For production work with broader kernels, consider `scipy.signal.convolve` with an explicit boundary mode such as `'reflect'` or `'nearest'`.\n",
+    "```"
```
damskii9992: Maybe add a comment that this is a notebook to show of Bayesian analysis, and urge users to check out `EasyDynamics` TM for better QENS models?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "\n",
+    "\n",
+    "def lorentzian(x: np.ndarray) -> np.ndarray:\n",
+    "    return A.value / np.pi * gamma.value / ((x - omega_0.value) ** 2 + gamma.value**2)\n",
+    "\n",
+    "\n",
+    "def intensity(x: np.ndarray) -> np.ndarray:\n",
+    "    gauss = norm(0, sigma.value).pdf(x)\n",
+    "    gauss /= gauss.sum()\n",
+    "    return np.convolve(lorentzian(x), gauss, 'same')"
```
damskii9992: I would swap the order of cells here. It is weird to define a function with arguments that does not yet exist . . .
I know it makes more sense to define the model first and THEN set the parameters, but from a coding-point of view this is weird. The editors also complain because the parameters haven't been defined yet . . .

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "Create four `Parameter` objects, for $A$, $\\gamma$, $\\omega_0$ and $\\sigma$. The `min` and `max` arguments define a **uniform prior** on each parameter &mdash; the sampler will only consider values inside this range and will treat every value inside the range as equally plausible *a priori*.\n",
+    "\n",
+    "| Parameter | Initial Value | Min | Max |\n",
+    "| --- | --- | --- | --- |\n",
+    "| $A$ | 10 | 1 | 100 |\n",
+    "| $\\gamma$ | 8.0 &times; 10<sup>-3</sup> | 1.0 &times; 10<sup>-4</sup> | 1.0 &times; 10<sup>-2</sup> |\n",
+    "| $\\omega_0$ | 1.0 &times; 10<sup>-3</sup> | 0 | 2.0 &times; 10<sup>-3</sup> |\n",
+    "| $\\sigma$ | 1.0 &times; 10<sup>-3</sup> | 1.0 &times; 10<sup>-5</sup> | 1.0 &times; 10<sup>-1</sup> |\n",
+    "\n",
+    "**Solution:**"
```
damskii9992: This is not the summer school, its a tutorial, we don't do exercises with solutions, we simply show-case the software.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "10",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from easyscience import Parameter\n",
+    "\n",
+    "A = Parameter(name='A', value=10, fixed=False, min=1, max=100)\n",
```
damskii9992: Call it "amplitude", please.

damskii9992:
> It's using `ObjBase`, which we're deprecating

Sadly we don't have the replacement yet. We need a new class for these kinds of `easyscience` standalone fits where you can just feed the class the parameters and the model function. To make it easy. Using `ModelBase` requires defining a class which is not something a user should be doing . . . .

So I think we should stick with using `ObjBase` (at least in tutorials) until we have created the replacement class.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "        self._omega_0.value = value\n",
+    "\n",
+    "    @property\n",
+    "    def sigma(self):\n",
+    "        return self._sigma\n",
+    "\n",
+    "    @sigma.setter\n",
+    "    def sigma(self, value):\n",
+    "        self._sigma.value = value\n",
+    "\n",
```
damskii9992: See my other comment regarding the use of `ModelBase`.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "\n",
+    "params = QENSModel(A=A, gamma=gamma, omega_0=omega_0, sigma=sigma)\n",
+    "\n",
+    "mle_fitter = Fitter(params, intensity)\n",
+    "mle_result = mle_fitter.fit(x=omega, y=intensity_obs, weights=1 / di)\n",
+    "\n",
+    "A, gamma, omega_0, sigma"
```
damskii9992: Put print statements on these. To need to have the output be a tuple.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "13",
+   "metadata": {},
+   "source": [
+    "## Exercise 3: switch to BUMPS and draw posterior samples\n",
+    "\n",
+    "`MultiFitter.sample` exposes the BUMPS DREAM sampler. It accepts a list of datasets (one entry per model) and returns a dictionary with the following keys:\n",
```
damskii9992: Could we give the method a more appropriate name? `bayesian_sampler`, or `mcmc_sample`? Simply calling it `sample` makes me thing of the `sample_model`.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "14",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from easyscience import AvailableMinimizers\n",
+    "from easyscience.fitting.multi_fitter import MultiFitter\n",
+    "\n",
+    "sampler = MultiFitter([params], [intensity])\n",
```
damskii9992: Why are we using the `MultiFitter` for this? Can't this be done with the regular `Fitter`?
It seems weird to use the `MultiFitter` with for a single model/dataset . . .

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "outputs": [],
+   "source": [
+    "from scipy.stats import norm\n",
+    "\n",
+    "\n",
+    "def lorentzian(x: np.ndarray) -> np.ndarray:\n",
+    "    return A.value / np.pi * gamma.value / ((x - omega_0.value) ** 2 + gamma.value**2)\n",
+    "\n",
+    "\n",
+    "def intensity(x: np.ndarray) -> np.ndarray:\n",
```
damskii9992: Since this is the model, please give the function a more appropriate name, like: `intensity_model`

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "execution_count": null,
+   "id": "12",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from easyscience import Fitter\n",
+    "from easyscience.base_classes import ModelBase\n",
+    "\n",
+    "\n",
+    "class QENSModel(ModelBase):\n",
```
damskii9992: Since this is just a container for the parameters, please give it an appropriate name, like `parameter_container`. It is not the model.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "sampler.switch_minimizer(AvailableMinimizers.Bumps)\n",
+    "\n",
+    "result = sampler.sample(\n",
+    "    x=[omega],\n",
+    "    y=[intensity_obs],\n",
+    "    weights=[1 / di],\n",
+    "    samples=4000,\n",
+    "    burn=500,\n",
+    "    thin=2,\n",
+    "    seed=42,\n",
```
damskii9992: You're giving this function a lot of inputs without explaining what they do. This is a tutorial on how to do Bayesian sampling, please explain what these arguments are and why you choose these values.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "13",
+   "metadata": {},
+   "source": [
+    "## Exercise 3: switch to BUMPS and draw posterior samples\n",
+    "\n",
+    "`MultiFitter.sample` exposes the BUMPS DREAM sampler. It accepts a list of datasets (one entry per model) and returns a dictionary with the following keys:\n",
+    "\n",
+    "- `draws`: a `(n_samples, n_parameters)` array of posterior samples;\n",
```
damskii9992: Are we drawing `n_samples` for each of `n_parameters`?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "\n",
+    "A, gamma, omega_0, sigma"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "13",
+   "metadata": {},
+   "source": [
+    "## Exercise 3: switch to BUMPS and draw posterior samples\n",
```
damskii9992: Please explain in at least a little detail what we actually do here. I.e. something like:
"We now make 4000 random draws from the calculated posterier distributions" (I am not sure myself what we are doing here, but please explain it, or reference where it is explained in another document).

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from easyscience import AvailableMinimizers\n",
+    "from easyscience.fitting.multi_fitter import MultiFitter\n",
+    "\n",
+    "sampler = MultiFitter([params], [intensity])\n",
+    "sampler.switch_minimizer(AvailableMinimizers.Bumps)\n",
+    "\n",
+    "result = sampler.sample(\n",
```
damskii9992: I would probably split this cell into two cells. The first where you initiate the `MultiFitter` and switch to `Bumps`, and the other where you actually run the sampler. Running calculations should really be their own cells always.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "cell_type": "markdown",
+   "id": "8766b170",
+   "metadata": {},
+   "source": [
+    "## Exercise 4: convergence diagnostics\n",
+    "\n",
+    "Before we trust the posterior samples, we should check that the MCMC chains have **converged** &mdash; that is, the sampler has found the typical set of the posterior and is no longer drifting. Two simple visual checks are:\n",
+    "\n",
+    "1. **Trace plot** &mdash; plot the sampled parameter values against the sample index. A well-converged chain looks like a \"hairy caterpillar\": it fluctuates around a stable mean with no long-term trends.\n",
+    "2. **Log-posterior plot** &mdash; the log-posterior $\\log p(\\theta \\mid d)$ should also stabilise after burn-in. If it is still climbing at the end of the run, the sampler has not yet converged.\n",
```
damskii9992, line 311: Heh, not the most scientific description of "it should only show the expected random fluctuations", but I kinda like it.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "\n",
+    "The `logp` array returned by `sample` contains the log-posterior for every *retained* sample (i.e. *after* burn-in and thinning). This means a flat `logp` trace is exactly what we want to see.\n",
```
damskii9992: You have not explained what burn-in and thinning is. But maybe you will have addressed this by now following my earlier comment.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "    weights=[1 / di],\n",
+    "    samples=4000,\n",
+    "    burn=500,\n",
+    "    thin=2,\n",
+    "    seed=42,\n",
+    ")\n",
+    "\n",
+    "print(f'Drew {result[\"draws\"].shape[0]} samples for {result[\"draws\"].shape[1]} parameters.')\n",
+    "print('parameters:', result['param_names'])"
+   ]
```
damskii9992, line 299: I'm getting a warning: Did not converge! Here

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "    ax.set_xlim(0, len(draws) - 1)\n",
+    "\n",
+    "# Log-posterior trace\n",
+    "axes[4].plot(logp, lw=0.5, color='C4')\n",
+    "axes[4].set_ylabel('log-posterior')\n",
+    "axes[4].set_xlabel('sample index')\n",
+    "\n",
+    "fig.suptitle('MCMC trace plots — check for \"hairy caterpillar\" behaviour')\n",
+    "fig.tight_layout()\n",
+    "plt.show()"
```
damskii9992, line 353: This functionality should probably be a part of the Bayesian implementation, rather than a function in the tutorial notebook. It should be a simple call like: `result.plot_convergence()`

damskii9992: I realize this means that the `results` output need to be a kind of results class, rather than just a dictionary.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "source": [
+    "summary_rows = []\n",
+    "for label, par in (('A', A), ('gamma', gamma), ('omega_0', omega_0), ('sigma', sigma)):\n",
+    "    col = column_for(par)\n",
+    "    lo, med, hi = np.percentile(col, [16, 50, 84])\n",
+    "    summary_rows.append((label, med, med - lo, hi - med, par.value))\n",
+    "\n",
+    "print(f'{\"param\":<8s} {\"median\":>12s} {\"-1σ\":>12s} {\"+1σ\":>12s} {\"MLE\":>12s}')\n",
+    "for label, med, minus, plus, mle in summary_rows:\n",
+    "    print(f'{label:<8s} {med:12.4g} {minus:12.4g} {plus:12.4g} {mle:12.4g}')"
```
damskii9992: You're tabulating the median and 16/84th percentiles, not +- 1 standard distribution, so don't use sigma for the table headers. Use 16 and 84th percentile.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "17",
+   "metadata": {},
+   "source": [
+    "## Exercise 6: visualise the joint posterior\n",
+    "\n",
+    "Marginal summaries hide correlations between parameters. A *corner plot* (a triangular grid of pairwise scatter plots and 1-D histograms) is the standard way to display them. Below we build one with plain `matplotlib`; if you have the [`corner`](https://corner.readthedocs.io/) package installed it will produce a publication-quality version with a single call.\n",
```
damskii9992: Instead of: "if you have the `corner` package installed it will produce a publication-quality" etc
Write: "To produce a better quality plot, use plotting packages such as `corner` which can produce publication-quality corner plots for you"
Otherwise you make it sound like the user should already have `corner` installed for some reason, or that the following cell produces this nice plot for you if you have this package installed (it does not).

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "        if i == n - 1:\n",
+    "            ax.set_xlabel(labels[j])\n",
+    "        else:\n",
+    "            ax.set_xticklabels([])\n",
+    "        if j == 0 and i != 0:\n",
+    "            ax.set_ylabel(labels[i])\n",
+    "        else:\n",
+    "            ax.set_yticklabels([])\n",
+    "fig.tight_layout()\n",
+    "plt.show()"
```
damskii9992, line 446: The top-left histogram plot is missing its y-label of "A" (or Amplitude, if you changed it).

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "19",
+   "metadata": {},
+   "source": [
+    "## Exercise 7: posterior-predictive band\n",
+    "\n",
+    "One of the practical benefits of having the full posterior in hand is that it is trivial to propagate the uncertainty in the parameters through to predictions of the model. We pick a few hundred random draws, evaluate the model at each, and plot the resulting envelope alongside the data.\n",
```
damskii9992: Can't you do exactly the same with a maximum likelihood estimate? Just randomly draw from the resulting Gaussian distributions and evaluate the model at each draw to plot the envelope.
I don't really see how this is any different with Bayesian analysis.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "hi = np.percentile(predictions, 97.5, axis=0)\n",
+    "mid = np.percentile(predictions, 50, axis=0)\n",
+    "\n",
+    "fig, ax = plt.subplots()\n",
+    "ax.errorbar(omega, intensity_obs, di, fmt='.', label='data')\n",
+    "ax.fill_between(omega, lo, hi, color='C1', alpha=0.3, label='95% posterior band')\n",
+    "ax.plot(omega, mid, '-', color='C1', label='posterior median')\n",
+    "ax.set(xlabel='$\\\\omega$/meV', ylabel='$I(\\\\omega)$')\n",
+    "ax.legend()\n",
+    "plt.show()"
```
damskii9992, line 493: This should really be hidden away by a method in `easyscience` to make it a simple call like:
```python
results.plot_model(confidence_intervals = [2.5, 97.5])
```
Or something like that in terms if the user-facing API.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "metadata": {},
+   "source": [
+    "## What next?\n",
+    "\n",
+    "A few things to try as further exercises:\n",
+    "\n",
+    "- Tighten or widen the `min`/`max` bounds on one of the parameters and see how the posterior reacts. This is the simplest possible *prior-sensitivity* analysis.\n",
+    "- Use the `chains` argument to `sample` to increase the DREAM population, e.g. `chains=10`, and check that the marginal histograms and trace plots are unchanged &mdash; a sign that the chains are sampling the same distribution.\n",
+    "- Pass a `progress_callback` to `sample` to monitor the sampler in real time &mdash; the same callback protocol used by the classical fitters works during MCMC (with `payload['sampling']` set to `True`).\n",
+    "- Inspect `result['state']` directly; the BUMPS [`MCMCDraw`](https://bumps.readthedocs.io/en/latest/api/bumps.dream.html) object exposes additional diagnostics such as per-chain log-posteriors, autocorrelation estimates, and the Gelman-Rubin $\\hat{R}$ statistic for formal convergence assessment."
```
damskii9992: This is not a summer school exercise. I would not include these things in a tutorial.
Maybe the inspect `result['state']` with a link to the documentation for the API in Bumps, but not the rest.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -0,0 +1,554 @@
...
+   "id": "13",
+   "metadata": {},
+   "source": [
+    "## Exercise 3: switch to BUMPS and draw posterior samples\n",
+    "\n",
+    "`MultiFitter.sample` exposes the BUMPS DREAM sampler. It accepts a list of datasets (one entry per model) and returns a dictionary with the following keys:\n",
+    "\n",
+    "- `draws`: a `(n_samples, n_parameters)` array of posterior samples;\n",
+    "- `param_names`: the unique names of the parameters, in the same column order as `draws`;\n",
+    "- `state`: the underlying BUMPS `MCMCDraw` object, useful for advanced diagnostics;\n",
```
damskii9992: I wouldn't call this "state" but rather "internal_bumps_object" or something similar and more descriptive.

damskii9992 (commented):
I have only reviewed the notebook so far, but I feel like there is enough here to work on while I review the code.

docs/mkdocs.yml
```diff
@@ -179,6 +179,7 @@ nav:
       - Workshops & Schools:
           - Fitting QENS: tutorials/fitting-qens.ipynb
           - Fitting SANS: tutorials/fitting-sans.ipynb
+          - Bayesian analysis: tutorials/fitting-bayesian.ipynb
```
damskii9992: Why is this under Workshops & Schools? Isn't this a regular tutorial?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+            User-friendly alias for BUMPS DREAM population count.
+        population : int | None, default=None
+            BUMPS DREAM population count for advanced users.
+        seed : int | None, default=None
+            Best-effort random seed.  Calls ``numpy.random.seed(seed)``
+            before DREAM starts, which affects the *global* NumPy RNG
+            state and may interact with other code in the process.
+            BUMPS DREAM uses additional internal RNG state that is
+            **not** controlled by this seed, so exact reproducibility
+            across runs is **not** guaranteed.
```
damskii9992: Wait, if BUMPS have an internal RNG state where we can't choose to the seed, then I don't think it makes sense to even have an argument for it.
If I see this argument, I would fully believe that setting it guarantees reproducability, so just having this argument confuses users, since it doesn't do what you'd think it does.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+        weights : np.ndarray
+            Flattened weight array.
+        samples : int, default=10000
+            Number of retained DREAM samples requested from BUMPS.
+        burn : int, default=2000
+            Burn-in steps.
+        thin : int, default=10
+            Thinning interval.
+        chains : int | None, default=None
+            User-friendly alias for BUMPS DREAM population count.
```
damskii9992: Why?
I don't particularly think that `chains` is a more user-friendly argument name than `population`. And we definitely shouldn't support both arguments as that just adds even more confusion.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+        FitError
+            If DREAM sampling was aborted by the user (via ``abort_test``).
+        Exception
+            Re-raised from DREAM fitting if any unexpected error occurs
+            (parameter values are restored beforehand).
+        """
+        from bumps.fitters import DreamFit
+        from bumps.names import FitProblem
+
+        x, y, weights = np.asarray(x), np.asarray(y), np.asarray(weights)
```
damskii9992, line 451: No check that the input is a list or something that can actually be converted to numpy arrays?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+
+        if y.shape != x.shape:
+            raise ValueError('x and y must have the same shape.')
+
+        if weights.shape != x.shape:
+            raise ValueError('Weights must have the same shape as x and y.')
+
+        if not np.isfinite(weights).all():
+            raise ValueError('Weights cannot be NaN or infinite.')
```
damskii9992, line 472: Why are you only checking weights? Couldn't there also be NaN or infinites in the x or y?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+
+        if (weights <= 0).any():
+            raise ValueError('Weights must be strictly positive and non-zero.')
+
```
damskii9992, line 476: According to codecov, none of the above checks are being tested in unit tests.

damskii9992: Also, what about the remaining arguments? They haven't been checked.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+        # Resolve population parameter
+        pop = self._resolve_population_alias(chains, population)
+
+        # Build DREAM kwargs
+        dream_kwargs: dict = {'samples': samples, 'burn': burn, 'thin': thin}
+        if pop is not None:
+            dream_kwargs['pop'] = pop
+        if sampler_kwargs:
+            dream_kwargs.update(sampler_kwargs)
```
damskii9992: Several of the these lines will fail if the arguments haven't been checked to be of the correct types.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+
+        # Build monitors (same pattern as classical Bumps.fit())
+        monitors = []
+        if progress_callback is not None:
+            if not callable(progress_callback):
+                raise ValueError('progress_callback must be callable')
+            # Compute total DREAM steps for progress display (burn + sampling generations)
+            pop_val = pop if pop else 10
```
damskii9992: Please. you saved 2 letters to not write `pop_value` . . . .
Why 10 here? Is that BUMPS default? If it is, shouldn't you just set `pop` equal to that earlier?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -374,6 +417,186 @@ def _make_func(x, y, weights):
...
+                        'total_steps': _total_steps,
+                    },
+                )
+            )
+
+        driver = FitDriver(
+            fitclass=DreamFit,
+            problem=problem,
+            monitors=monitors,
+            abort_test=abort_test or (lambda: False),
```
damskii9992: Huh? Why the or statement here?

src/easyscience/fitting/multi_fitter.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from typing import Callable
+from typing import Dict
```
damskii9992: Wasn't it decided that we would use the new type-hinting syntax? This syntax was deprecated in Python 3.9

src/easyscience/fitting/multi_fitter.py
```diff
@@ -188,3 +189,122 @@ def _post_compute_reshaping(
...
+                population=pop,
+                seed=seed,
+                sampler_kwargs=sampler_kwargs,
+                progress_callback=progress_callback,
+                abort_test=abort_test,
+            )
+        finally:
+            self.fit_function = original_fit_func
+
+        return result
```
damskii9992: All my comments obviously carry over to the MultiFitter method too.

tests/integration/fitting/test_multi_fitter.py
```diff
@@ -286,3 +286,303 @@ def test_multi_fit_1D_2D(fit_engine):
...
+            y=[y],
+            weights=[weights],
+            samples=100,
+            burn=20,
+            thin=2,
+            sampler_kwargs={'init': 'random'},
+        )
+
+        assert result['draws'].ndim == 2
+        assert result['draws'].shape[0] > 0
```
damskii9992: Didn't we decide to have 1 test class for 1 source class? So shouldn't these all be consolidated to a single class? You're only using multiple classes to group different tests. You can do that by simply adding comment headers like the one you have at the top.

tests/integration/fitting/test_multi_fitter.py
```diff
@@ -286,3 +286,303 @@ def test_multi_fit_1D_2D(fit_engine):
...
+        sp.phase.fixed = False
+
+        f = MultiFitter([sp], [sp])
+        try:
+            f.switch_minimizer('Bumps')
+        except AttributeError:
+            pytest.skip('BUMPS is not installed')
+
+        result = f.sample(
+            x=[x2D], y=[y2D], weights=[weights], samples=100, burn=20, thin=2, vectorized=True
```
damskii9992, line 424: What is this `vectorized` keyword? And since we don't expose it in our API, why do we even test for it?

tests/integration/fitting/test_multi_fitter.py
damskii9992: This obviously also count for the minimizer_bumps.py tests too.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Then we need to apply this to all the other notebooks. If we decide to do so, let's make it a separate issue and address it there.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: You mean relying on `dmsc-summer-school` location or using `pooch.retrieve`?

tests/integration/fitting/test_multi_fitter.py
damskii9992: Another comment is that many of these tests aren't really integration tests, but more like unit tests. In my opinion.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: I wanted to leave the plotting implementation to the consumer. We shouldn't really enforce any specific plotting mechanism, only provide data.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: I'd add it to the EasyScience documentation issue, since it's going to be generated automatically, anyway.

pyproject.toml
```diff
@@ -211,8 +211,7 @@ select = [
 # Ignore specific rules globally
 ignore = [
   'COM812', # https://docs.astral.sh/ruff/rules/missing-trailing-comma/
-  # The following is replaced by 'D'/[tool.ruff.lint.pydocstyle] and [tool.pydoclint]
-  'DOC', # https://docs.astral.sh/ruff/rules/#pydoclint-doc
+  # The following is replaced by 'D'/[tool.ruff.lint.pydocstyle] and [tool.pydoclint]  'DOC', # https://docs.astral.sh/ruff/rules/#pydoclint-doc
```
damskii9992: Why was this commented out?

damskii9992 (commented):
Comments after reading the code and not just the user-facing API.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: I'll leave it since it is descriptive, and Henrik didn't complain :)

docs/docs/tutorials/fitting-bayesian.ipynb
henrikjacobsenfys: It's not the amplitude of the function, though, it's the area

src/easyscience/fitting/multi_fitter.py
henrikjacobsenfys: I think this can be auto-fixed by Ruff

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Results being a class limit the applicability of the code. The consumer needs to know the class layout and adhere to it. Using a simple dict is far simpler and far superior, I would claim.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: As mentioned above - we don't want the core to provide us with plotting. Users choose what plotting mechanism they use with the data they get from fitting. Hence - no plopp/matplotlib/plotly dependence in the core itself.

tests/integration/fitting/test_multi_fitter.py
rozyczko: `vectorized` is indeed in `mcmc_sample` signature. It is quite important so I'll leave it as is

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: The lambda is just an no-op. The driver calls it periodically during the fit loop, and as long as it returns False the fit continues. If a real abort_test is provided and returns True, the driver stops early.
`abort_test` must be a callable, not a boolean, hence `lambda`

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: removed `chains`.  It was sort of useful for non-BUMPS people, but let's have it unequivocal.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Still I think it is useful to add the seed, for possible reproducibility. I added a comment but don't really want to remove it

docs/mkdocs.yml
rozyczko: moved up

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: removed What's Next entirely. No need.

damskii9992:
> > It's using `ObjBase`, which we're deprecating
> 
> Sadly we don't have the replacement yet. We need a new class for these kinds of `easyscience` standalone fits where you can just feed the class the parameters and the model function. To make it easy. Using `ModelBase` requires defining a class which is not something a user should be doing . . . .
> 
> So I think we should stick with using `ObjBase` (at least in tutorials) until we have created the replacement class.

Actually, come to think of it, I believe we can replace it with an `EasyList` for now. All the class needs to do is hold the parameters and have an `get_all_fittable_parameters` method implemented. I think that would be cleaner.

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: I mean something like what I have in my documentation EasyImaging branch.

A `datasets` module with loader functions, which loads in files from its own pooch registry.

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: Do we have an issue on that yet? 👀

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: Well, you just proved my point then XD "A" is too vague :P

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: added description in the cell above

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: Hmm, in our own technique libraries we very much enforce specific plotting mechanisms, but this is the core `easyscience` library and I guess direct users of `easyscience` are not expected to be coding impaired. 
I would still have some convenience plotting mechanisms implemented, maybe just methods from a `plotting` submodule? Since many users will not want to make their own plots and will prefer the convenience of pre-made plotting functions.

rozyczko: This means either asking users to install easycore with `.[dev]` or add plotting dependences to the main list. This is why scipp spun off plopp, so the core module didn't have to depend on plotting libraries.

We didn't require plotting to be installed for viewing fitting results, so why do you think it is necessary for the sampling results? The result dictionary from both fitting and sampling is available for plotting anyway.

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992: But if it doesn't actually provide reproducibility, then what is the point of it? Semi-reproducibility seems to me to be self-contradictory.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: converted to `abort_test=abort_test if abort_test is not None else (lambda: False)`

tests/integration/fitting/test_multi_fitter.py
damskii9992: Can you then add a comment to the test to explain why/what? And should it be exposed in the API?

rozyczko:
> > > It's using `ObjBase`, which we're deprecating
> > 
> > 
> > Sadly we don't have the replacement yet. We need a new class for these kinds of `easyscience` standalone fits where you can just feed the class the parameters and the model function. To make it easy. Using `ModelBase` requires defining a class which is not something a user should be doing . . . .
> > So I think we should stick with using `ObjBase` (at least in tutorials) until we have created the replacement class.
> 
> Actually, come to think of it, I believe we can replace it with an `EasyList` for now. All the class needs to do is hold the parameters and have an `get_all_fittable_parameters` method implemented. I think that would be cleaner.

As discussed, I'll move back to `ObjBase` so all the tutorials can be modified to use EasyList at the same time, in a separate PR

https://github.com/easyscience/core/issues/253

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: https://github.com/easyscience/core/issues/254

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: https://github.com/easyscience/core/issues/255

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Modified the values so it converges. Runs for way longer but it is necessary.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Also, https://github.com/easyscience/core/issues/255

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: This, again, is a good refactoring, which needs to be done for all the tutorials, hence outside the scope of this PR. Maybe we can move it out to EasyUtils

https://github.com/easyscience/core/issues/256

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Fine. Removed. This might indeed be a bit misleading if users rely on it.

tests/integration/fitting/test_multi_fitter.py
rozyczko: It is not just in sampling but in regular fitting as well. Nothing new.
But I added more comments explaining the usage cases.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Because I'm a numpty. Fixed

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: > We didn't require plotting to be installed for viewing fitting results, so why do you think it is necessary for the sampling results? The result dictionary from both fitting and sampling is available for plotting anyway.

Because the results from fitting are more easily understandable without plotting. For the Bayesian sampling, the results pretty much require plotting to be viewed . . .
But maybe this is another good contender for `EasyUtils` . . .

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: @rozyczko As Henrik pointed out, "A" is apparently the area, so it should be renamed to "area" rather than "amplitude" ;)

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: I am still getting the warning :)

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Already in https://github.com/easyscience/core/issues/255

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: My changes to the number of iterations didn;'t get saved... my bad. Fixed

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Fixed

damskii9992 (approved):
LGTM for now.

PR #247, opened 2026-05-07
Apply latest Copier template updates

AndrewSazonov:
This PR syncs the project with the latest Copier templates. Main changes include replacing deprecated Node.js 20 usage, migrating docstrings from Google style to NumPy style, and applying related template-driven configuration updates.


henrikjacobsenfys (approved):
I looked at a handful of randomly selected files, and they looked fine.

rozyczko (approved):
I spent way too much time on this PR :)

Rechecked actions version bump correctness,
Assured `client-id` is the new version of `app-ad` in `create-github-app-token` action
Checked dependency move in pyproject.toml
Got worried about manually copying `trstringer/require-label-prefix` action
Very useful addition of `nodefaults` to pixi channels
etc.

One potential issue I have is - do we really, REALLY, need to add the useless docstrings to all the methods?
```python
def __init__(self):
    """Init function."""
```
```python
def __call__(self, *args, **kwargs):
    """Call function."""
```

Instead of adding junk, why not just `# noqa: DOC` comment for those trivial dunder methods?

comments

damskii9992:
> Got worried about manually copying `trstringer/require-label-prefix` action

Yeah, I am not a big fan of that either. But on the other hand, that action is currently not maintained. 
Hmm, I remember I saw a PR on the repo which got closed. Looking at the guy who posted the PR, I figured he might have made his own implementation, since `trstringer` is apparently MIA. And lo an behold, look at what I found:
https://github.com/Rindrics/expect-label-prefix
What do you think @AndrewSazonov ?

damskii9992:
> Very useful addition of `nodefaults` to pixi channels

I believe this was required for us now, right? To make 100% sure we don't accidentally get any commercial conda-packages.

src/easyscience/base_classes/easy_list.py
```diff
@@ -36,14 +36,28 @@ def __init__(
...
+        ----------
+        *args : ProtectedType_ | list[ProtectedType_]
+            Initial items to add to the list.
+        protected_types : list[Type[NewBase]] | Type[NewBase] | None, default=None
+            Types that are allowed in the list. Can be a single NewBase
+            subclass or a list of them. If None,. By default, None.
+        unique_name : Optional[str], default=None
+            Optional unique name for the list. By default, None.
+        display_name : Optional[str], default=None
+            Optional display name for the list. By default, None.
```
damskii9992, line 52: The automatic docstring conversion clearly took some liberties in regards to scanning scanning the code and re-writing the docstrings. We probably need to manually go through the docstrings after this and ammend them.

src/easyscience/legacy/collection_base.py
```diff
@@ -25,16 +25,18 @@
```
damskii9992, line 25: We should probably make everything ignore files in the legacy folder. Including coverage, linting, docstrings etc. The legacy folder isn't supported and thus shouldn't be maintained. IMO.

pixi.toml
```diff
@@ -1,72 +1,61 @@
...
+
+# User-like behavior for testing with pip-installed dependencies instead
+# of editable installations.
+
+[feature.user.dependencies]
+pixi-kernel = '*' # Pixi Jupyter kernel
+
+[feature.user.pypi-dependencies]
+pip = '*'
+easydiffraction = '*'
```
damskii9992: ??? easydiffraction?

pyproject.toml
```diff
@@ -23,25 +23,24 @@ classifiers = [
...
+  'lmfit',   # Non-linear least squares fitting
+  'bumps',   # Bayesian uncertainty estimation
+  'dfo-ls',  # Derivative-free optimization
+  'numpy',   # Numerical computing
+  'scipp',   # Handling and analysis of scientific data
 ]
 
 [project.optional-dependencies]
 dev = [
+  'uncertainties',                   # Propagation of uncertainties in calculations
```
damskii9992: I think we can remove this dependency completely, also from dev.

pyproject.toml
```diff
@@ -198,46 +207,96 @@ select = [
...
+'docs/**' = [
+  'INP001', # https://docs.astral.sh/ruff/rules/implicit-namespace-package/
+  'T201',   # https://docs.astral.sh/ruff/rules/print/
+]
+
+# Specific options for certain rules
 
 [tool.ruff.lint.flake8-tidy-imports]
+# Disallow all relative imports
 ban-relative-imports = 'all'
```
damskii9992, line 241: Didn't we discuss that this was a developer preference? I for one prefer relative imports over absolute imports.

rozyczko:
The one blocking issue is the absolute path fix in `test.yml`. Once this is sorted out, we should merge this PR ASAP

pyproject.toml
AndrewSazonov: Right, we did discuss this. Personally, I do not mind whether imports are absolute or relative, but I do care about consistency.

It seems that the only way to guarantee consistency is to enforce absolute imports everywhere. Otherwise, we end up with a random mix of both styles in different places.

Therefore, I prefer absolute imports. That said, if the majority prefers relative imports, I am okay with removing this setting.

src/easyscience/base_classes/easy_list.py
AndrewSazonov: I agree that a manual check is needed, and that we should all go through the docstrings carefully. But do you suggest doing this as part of the current PR, or plan it as a follow-up task during future development?

AndrewSazonov:
> One potential issue I have is - do we really, REALLY, need to add the useless docstrings to all the methods?
> Instead of adding junk, why not just # noqa: DOC comment for those trivial dunder methods?

Agreed, these placeholder docstrings are not useful.

I also prefer not to add  `# noqa` comments, since that would add a different kind of noise. I checked the current docstring tooling, and it does not require every method to have a docstring: `pydoclint` validates existing docstrings, and `format-docstring` formats existing docstrings.

So I removed the useless placeholder docstrings instead. `interrogate` is still active and reports docstring coverage, so we keep visibility into the current documentation level.

AndrewSazonov:
> > Got worried about manually copying `trstringer/require-label-prefix` action
> 
> Yeah, I am not a big fan of that either. But on the other hand, that action is currently not maintained. Hmm, I remember I saw a PR on the repo which got closed. Looking at the guy who posted the PR, I figured he might have made his own implementation, since `trstringer` is apparently MIA. And lo an behold, look at what I found: https://github.com/Rindrics/expect-label-prefix What do you think @AndrewSazonov ?

Thanks for finding that. I switched the templates from the custom script to `Rindrics/expect-label-prefix` as suggested, and then reapplied the updated templates here.

AndrewSazonov:
> The one blocking issue is the absolute path fix in `test.yml`. Once this is sorted out, we should merge this PR ASAP

This issue is resolved now. Thanks for suggesting the fix.

pyproject.toml
rozyczko: Or maybe add this as a question at the template generation stage? Default to relative but allow override.
Would that be a solution?

AndrewSazonov: This can of course be added and configured differently for different technique-specific projects. But I think the main question here is which style we want to use in the `easyscience` module.

src/easyscience/base_classes/easy_list.py
damskii9992: Definitely a follow-up task for a future documentation effort :)

damskii9992 (approved):
Yeah, I think this is ready to be merged now.

pyproject.toml
AndrewSazonov: Added feature request: easyscience/templates#35

PR #248, opened 2026-05-12
Bump max Python to 3.14 and update dependencies

AndrewSazonov:


rozyczko (approved):
Looks good to both me and DeepSeek.

PR #249, opened 2026-05-13
src\legacy should be in wheel

rozyczko:
legacy projects still depend on components in src\easyscience\legacy


AndrewSazonov (approved):
LGTM

PR #252, opened 2026-05-21
Improved logging

### LINKED ISSUE #234: Make logging in core user controlled
Core logs a lot of `warning`s to STDOUT. Those should be user controllable.
If a technique-specific library doesn't want the core to pollute standard output, there should be a way to switch those logging messages off.

rozyczko:
Switched from direct `warnings.warn` and `print` statements to Python's standard `logging` module throughout the codebase. 

* Replaced all uses of `warnings.warn` and `print` for runtime and deprecation warnings with appropriate calls to the `logging` module, using named loggers under the `easyscience` hierarchy for better control and filtering. 

* Added a new user guide page, `controlling-log-output.md`, detailing how to suppress, filter, or redirect EasyScience log messages. 
 
* No global reconfiguration, no default stream handlers, and all loggers inherit configuration, giving users and downstream libraries full control over log output.

### DIFF
--- a/docs/docs/user-guide/controlling-log-output.md
+++ b/docs/docs/user-guide/controlling-log-output.md
@@ -1,0 +1,200 @@
+# Controlling Log Output
+
+EasyScience uses Python's standard `logging` module for all
+informational and warning messages. This gives you full control over
+what output the library produces and where it goes.
+
+## Quick Start
+
+The most common operation is suppressing all EasyScience messages. You
+can do it in one line:
+
+```python
+from easyscience import global_object
+
+global_object.log.set_level('ERROR')
+```
+
+Or using the standard library directly:
+
+```python
+import logging
+
+logging.getLogger('easyscience').setLevel(logging.ERROR)
+```
+
+Both forms set the package-root logger to `ERROR`, which suppresses all
+`WARNING`, `INFO`, and `DEBUG` messages from EasyScience.
+
+## Logger Hierarchy
+
+EasyScience loggers are named hierarchically under the root
+`easyscience` logger. This lets you suppress the whole package or
+individual subsystems:
+
+```
+easyscience                        # root — controls everything
+├── easyscience.base_classes       # EasyList runtime warnings
+├── easyscience.legacy             # ObjBase / CollectionBase deprecations
+├── easyscience.deprecated         # @deprecated decorator messages
+├── easyscience.fitting            # import-availability warnings
+│   ├── easyscience.fitting.bumps  # Bumps fitting runtime messages
+│   ├── easyscience.fitting.lmfit  # LMFit fitting runtime messages
+│   └── easyscience.fitting.dfo    # DFO fitting runtime messages
+├── easyscience.variable           # parameter warnings
+└── easyscience.global_object      # undo/redo, debugging diagnostics
+```
+
+Setting the level on a parent logger affects all its children, because
+child loggers inherit the parent's effective level.
+
+```python
+import logging
+
+# Suppress everything from the fitting subsystem (including bumps/lmfit/dfo)
+logging.getLogger('easyscience.fitting').setLevel(logging.ERROR)
+
+# Suppress only Bumps runtime messages
+logging.getLogger('easyscience.fitting.bumps').setLevel(logging.ERROR)
+
+# Suppress only legacy deprecation notices
+logging.getLogger('easyscience.legacy').setLevel(logging.ERROR)
+```
+
+## Environment Variable
+
+You can control the log level **before** importing EasyScience by
+setting the `EASYSCIENCE_LOG_LEVEL` environment variable. This is useful
+when a downstream library cannot configure logging in code before
+EasyScience is imported.
+
+**Accepted values:** `DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`, or
+a numeric level (e.g. `40`).
+
+**Example — suppressing all import-time messages:**
+
+```powershell
+# Windows PowerShell
+$env:EASYSCIENCE_LOG_LEVEL = 'ERROR'
+python -c "import easyscience"  # silent
+```
+
+```bash
+# Linux / macOS
+EASYSCIENCE_LOG_LEVEL=ERROR python -c "import easyscience"  # silent
+```
+
+Or from Python, set the environment variable before the import:
+
+```python
+import os
+
+os.environ['EASYSCIENCE_LOG_LEVEL'] = 'ERROR'
+import easyscience  # now silent
+```
+
+## Temporary Suppression (Context Manager)
+
+When you only want to silence messages during a specific operation, use
+the `at_level` context manager:
+
+```python
+from easyscience import global_object
+import logging
+
+# Messages during the fit are suppressed, then the previous level is restored
+with global_object.log.at_level(logging.ERROR):
+    results = fitter.fit(x, y, weights)
+```
+
+This is the recommended pattern for technique-specific libraries that
+don't want EasyScience output to leak into their own users' consoles.
+
+## Convenience API
+
+The `global_object.log` object exposes convenience methods that mirror
+`logging` module-level functions:
+
+| Method              | Description                                                          |
+| ------------------- | -------------------------------------------------------------------- |
+| `.set_level(level)` | Set the package-root logger level (`'WARNING'` or `logging.WARNING`) |
+| `.debug(msg)`       | Log a `DEBUG`-level message                                          |
+| `.info(msg)`        | Log an `INFO`-level message                                          |
+| `.warning(msg)`     | Log a `WARNING`-level message                                        |
+| `.error(msg)`       | Log an `ERROR`-level message                                         |
+| `.critical(msg)`    | Log a `CRITICAL`-level message                                       |
+| `.exception(msg)`   | Log an `ERROR`-level message with traceback                          |
+| `.getLogger(name)`  | Get a child logger under `easyscience`                               |
+| `.at_level(level)`  | Context manager for temporary level change                           |
+| `.suspend()`        | Suppress all EasyScience output                                      |
+| `.resume()`         | Restore the previously set level                                     |
+
+## Recipes
+
+### Recipe 1: Keep your test output clean
+
+When running tests that exercise EasyScience fitting, suppress the
+library's messages so they don't clutter your test reports:
+
+```python
+import logging
+from easyscience import global_object
+
+@pytest.fixture(autouse=True)
+def _quiet_easyscience():
+    with global_object.log.at_level(logging.ERROR):
+        yield
+```
+
+### Recipe 2: Library author embedding EasyScience
+
+If you maintain a library that uses EasyScience internally, prevent
+EasyScience from writing to your users' consoles:
+
+```python
+# In your library's __init__.py, before any EasyScience import:
+import os
+os.environ.setdefault('EASYSCIENCE_LOG_LEVEL', 'ERROR')
+```
+
+### Recipe 3: Debug mode for development
+
+When developing or debugging, see all EasyScience internal diagnostics:
+
+```python
+from easyscience import global_object
+import logging
+
+global_object.log.set_level('DEBUG')
+
+# Your code here — all EasyScience messages will be visible
+```
+
+### Recipe 4: See only error messages
+
+Keep the console clean but still see critical problems:
+
+```python
+from easyscience import global_object
+
+global_object.log.set_level('ERROR')
+```
+
+## Library-Safe Behaviour
+
+EasyScience follows standard library-logging best practices:
+
+- **No `logging.basicConfig()`** — EasyScience never reconfigures the
+  global logging setup.
+- **No default stream handlers** — EasyScience does not attach handlers
+  that write to `stdout` or `stderr`. It only creates log records.
+  Applications and test frameworks decide where those records go.
+- **Child loggers inherit** — Child loggers (e.g.
+  `easyscience.fitting.bumps`) are left at `logging.NOTSET` by default,
+  so they inherit level and handler configuration from the `easyscience`
+  package-root logger. Supressing the root suppresses everything.
+
+This means you are in full control. If you don't configure any handlers,
+EasyScience messages will not appear. If you do configure handlers (as
+`pytest` does via its logging plugin, or as an application might via
+`basicConfig`), you control what is shown and where.
--- a/docs/docs/user-guide/controlling-log-output.md
+++ b/docs/docs/user-guide/controlling-log-output.md
@@ -12,7 +12,7 @@ can do it in one line:
 ```python
 from easyscience import global_object
 
-global_object.log.set_level('ERROR')
+global_object.log.level = 'ERROR'
 ```
 
 Or using the standard library directly:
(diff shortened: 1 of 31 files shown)


docs/docs/user-guide/controlling-log-output.md
```diff
@@ -0,0 +1,200 @@
...
+library's messages so they don't clutter your test reports:
+
+```python
+import logging
+from easyscience import global_object
+
+@pytest.fixture(autouse=True)
+def _quiet_easyscience():
+    with global_object.log.at_level(logging.ERROR):
+        yield
```
damskii9992, line 147: How does this work? You're only muting the log in the context, so how would this actually mute the log in a test? Wouldn't
```python
global_object.log.set_level('ERROR')
```
Be better?

docs/docs/user-guide/controlling-log-output.md
```diff
@@ -0,0 +1,200 @@
...
+os.environ.setdefault('EASYSCIENCE_LOG_LEVEL', 'ERROR')
+```
+
+### Recipe 3: Debug mode for development
+
+When developing or debugging, see all EasyScience internal diagnostics:
+
+```python
+from easyscience import global_object
+import logging
```
damskii9992: You don't need to import logging, right?

docs/docs/user-guide/controlling-log-output.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Library-Safe Behaviour
+
+EasyScience follows standard library-logging best practices:
+
+- **No `logging.basicConfig()`** — EasyScience never reconfigures the
+  global logging setup.
+- **No default stream handlers** — EasyScience does not attach handlers
+  that write to `stdout` or `stderr`. It only creates log records.
+  Applications and test frameworks decide where those records go.
```
damskii9992, line 191: Isn't this bad? What about users that use `easyscience` as a standalone fitting package? Do they not see any messages without configuring it? They should . . .

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
 # SPDX-FileCopyrightText: 2026 EasyScience contributors <https://github.com/easyscience>
 # SPDX-License-Identifier: BSD-3-Clause
 
+import contextlib
 import logging
+import os
+from typing import Optional
```
damskii9992: Ehh, no Optional typing, use |

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+    """Parse an environment-variable string into a logging level."""
+    if raw is None:
+        return default
+    stripped = raw.strip()
+    if stripped.isdigit():
+        return int(stripped)
+    upper = stripped.upper()
+    if upper in _LEVEL_NAME_MAP:
+        return _LEVEL_NAME_MAP[upper]
+    return default
```
damskii9992: You should probably raise in the end. If you try to set a logging level, but make a spelling mistake, we shouldn't silently return the default.
Also, why the hell is the default a separate parameter? Why not just the default of `raw`? Which probably also should have a better parameter name (level?)

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
-        self.level = log_level
-        self.logger.setLevel(self.level)
+    """
+    Central logging controller for EasyScience.
+
+    Owns the package-root logger ``easyscience`` and provides a
+    convenience API to set its level.
+
+    Library-safe behaviour:
+    - Never calls :func:`logging.basicConfig`.
```
damskii9992: You didn't check the documentation, did you? :func: does not do anything in mkdocs. So you will just see the text ":func:", the same applied to other such similar calls, like :class: or :meth:.

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+    - Never attaches a default stream handler.
+    - Child loggers returned by :meth:`getLogger` are left at
+      ``logging.NOTSET`` so they inherit level control from the package
+      root logger.
```
damskii9992: This sounds like AI. Please remove it or re-write it to only provide the important information.

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+
+    Parameters
+    ----------
+    log_level : int, default=logging.WARNING
+        Default level for the package-root logger. Overridden by the
+        ``EASYSCIENCE_LOG_LEVEL`` environment variable when set.
```
damskii9992: This part of the docstring belongs to the `__init__`, as it is the __init__ which takes parameters.

src/easyscience/global_object/logger.py
damskii9992: Actually, this method probably shouldn't have any defaults . . .

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+    """
+
+    def __init__(self, log_level: int = logging.WARNING):
+        self._effective_default = _resolve_log_level(
+            os.environ.get(_LOG_LEVEL_ENV_VAR), default=log_level
+        )
+        self.logger = logging.getLogger(PACKAGE_LOGGER_NAME)
+        self.logger.setLevel(self._effective_default)
+        self.level = self._effective_default
```
damskii9992: Why are we keeping a `.level` attribute on the class? Can't we just make a property which gets the level from the `logging.logger`?

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+        """Log an ERROR-level message on the package-root logger."""
+        self.logger.error(msg, *args, **kwargs)
+
+    def critical(self, msg: str, *args, **kwargs) -> None:
+        """Log a CRITICAL-level message on the package-root logger."""
+        self.logger.critical(msg, *args, **kwargs)
+
+    def exception(self, msg: str, *args, **kwargs) -> None:
+        """Log an ERROR-level message with traceback on the package-root logger."""
+        self.logger.exception(msg, *args, **kwargs)
```
damskii9992, line 95: None of these seems to have unit tests. FYI. Not that I think this is important for these functions . . .

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
-        Create a logger :param color:.
+        Set the effective level of the package-root logger.
+
+        Parameters
+        ----------
+        level : int | str
+            Logging level — e.g. ``logging.WARNING``, ``'ERROR'``.
+        """
+        if isinstance(level, str):
+            level = _resolve_log_level(level)
```
damskii9992: don't overwrite the input parameter, call it `resolved_level` or `numeric_level` or something.

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+    def exception(self, msg: str, *args, **kwargs) -> None:
+        """Log an ERROR-level message with traceback on the package-root logger."""
+        self.logger.exception(msg, *args, **kwargs)
+
+    # -- public API -------------------------------------------------------
 
-    def getLogger(
-        self, logger_name: str, color: str = '32', defaults: bool = True
-    ) -> logging.Logger:
+    def set_level(self, level: int | str) -> None:
```
damskii9992: Could we not make this a property with a setter/getter method? Wouldn't that be cleaner? i.e.
```python
global_object.log.level = 'ERROR'
```
Instead of:
```python
global_object.log.set_level('ERROR')
```

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
 
         Returns
         -------
         logging.Logger
-            A logger.
+            A child logger whose name is ``easyscience.<logger_name>``
+            when *logger_name* does not already start with *easyscience*
+            or a dot.
+        """
+        if not logger_name.startswith('easyscience') and not logger_name.startswith('.'):
```
damskii9992: shouldnt `'easyscience'` here be `PACKAGE_LOGGER_NAME`?

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+            logger_name = f'{PACKAGE_LOGGER_NAME}.{logger_name}'
+        return logging.getLogger(logger_name)
+
+    @contextlib.contextmanager
+    def at_level(self, level: int | str):
+        """
+        Context manager that temporarily sets the package-root logger
+        to *level*, restoring the previous level on exit.
+
+        Example::
```
damskii9992: Use the numpy docs way of formatting examples. It's much prettier in the autogenerated mkdocs API reference:
```python
"""
    Example
    -------
    Setting the log-level only within the context:

    ```python
    with global_object.log.at_level(logging.ERROR):
       fitter.fit(x, y, weights)  # no core messages below ERROR
    ```
"""
```

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
-        #     for handler in self._handlers[handler_type]:
-        #         if handler_type == 'sys' or defaults:
-        #             handler.formatter._fmt = self._makeColorText(color)
-        #             logger.addHandler(handler)
-        # logger.propagate = False
-        # self._loggers.append(logger)
-        return logger
+        previous = self.level
+        if isinstance(level, str):
+            level = _resolve_log_level(level)
```
damskii9992: Same comment as above, don't overwrite the input.

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,161 @@
...
+
+    # -- deprecated helpers (compatibility shims) -------------------------
+
+    def suspend(self):
+        """Suppress all core log output (set level to CRITICAL+1)."""
+        self.logger.setLevel(logging.CRITICAL + 1)
+
+    def resume(self):
+        """Restore the previously configured log level."""
+        self.logger.setLevel(self.level)
```
damskii9992: I see here that making `level` a property would complicate this code a bit. But I still think it is worth it. You would just assign a temporary hidden variable to the old value when suspending, and then set the property to that value when resuming.

src/easyscience/base_classes/collection_base.py
```diff
@@ -7,13 +7,11 @@
...
 
 from ..legacy.collection_base import CollectionBase  # noqa: F401
 
-warnings.warn(
+logging.getLogger('easyscience').warning(
     'easyscience.base_classes.collection_base is deprecated. '
-    'Please import from easyscience.legacy.collection_base instead.',
-    DeprecationWarning,
-    stacklevel=2,
+    'Please import from easyscience.legacy.collection_base instead.'
```
damskii9992: Shouldn't this be using `global_object.log.warn` instead?
I think we should standardize on using the convenience method always, to ensure that we have full control over it in case we change something in the future.

src/easyscience/base_classes/collection_base.py
```diff
@@ -7,13 +7,11 @@
...
-import warnings
+import logging
 
 from ..legacy.collection_base import CollectionBase  # noqa: F401
 
-warnings.warn(
+logging.getLogger('easyscience').warning(
     'easyscience.base_classes.collection_base is deprecated. '
-    'Please import from easyscience.legacy.collection_base instead.',
-    DeprecationWarning,
```
damskii9992, line 17: So with the new logger, we lose the ability to categorize warnings?

src/easyscience/fitting/calculators/interface_factory.py
```diff
@@ -96,12 +97,16 @@ def switch(self, new_interface: str, fitter: Optional[Type[Fitter]] = None) -> N
                     if hasattr(obj, 'update_bindings'):
                         obj.update_bindings()
                 except Exception as e:
-                    print(f'Unable to auto generate bindings.\n{e}')
+                    logging.getLogger('easyscience.fitting.calculators').warning(
+                        'Unable to auto generate bindings.\n%s', e
+                    )
```
damskii9992, line 103: I assume you will ctrl+f to replace this everywhere, but just adding another comment here so you don't forget.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -485,18 +485,16 @@ def _gen_fit_results(
             )
         if stopped_on_budget:
             if tolerance is None:
-                warnings.warn(
+                logging.getLogger('easyscience.fitting.bumps').warning(
                     f'Fit did not converge within the maximum optimizer steps of {max_evaluations} '
                     f'({n_evaluations} objective evaluations). '
-                    'Consider increasing the maximum number of evaluations or adjusting the tolerance.',
-                    UserWarning,
```
damskii9992, line 492: Is it a problem that we lose the categories?

src/easyscience/fitting/available_minimizers.py
```diff
@@ -15,11 +15,8 @@
 
     lmfit_engine_available = True
 except ImportError:
-    # TODO make this a proper message (use logging?)
-    warnings.warn(
-        'LMFit minimization is not available. Probably lmfit has not been installed.',
-        ImportWarning,
```
damskii9992, line 21: Another category lost

src/easyscience/utils/decorators.py
```diff
@@ -99,11 +99,10 @@ def deprecated(func):
...
     def new_func(*args, **kwargs):
-        warnings.warn_explicit(
-            'Call to deprecated function {}.'.format(func.__name__),
-            category=DeprecationWarning,
-            filename=func.__code__.co_filename,
-            lineno=func.__code__.co_firstlineno + 1,
+        logging.getLogger('easyscience.deprecated').warning(
+            'Call to deprecated function %s.',
+            func.__name__,
+            stacklevel=3,
```
damskii9992, line 104: stacklevel? Is that a feature in the logging module? You haven't used this elsewhere.

pyproject.toml
```diff
@@ -211,8 +211,7 @@ select = [
 # Ignore specific rules globally
 ignore = [
   'COM812', # https://docs.astral.sh/ruff/rules/missing-trailing-comma/
-  # The following is replaced by 'D'/[tool.ruff.lint.pydocstyle] and [tool.pydoclint]
-  'DOC', # https://docs.astral.sh/ruff/rules/#pydoclint-doc
+  # The following is replaced by 'D'/[tool.ruff.lint.pydocstyle] and [tool.pydoclint]  'DOC', # https://docs.astral.sh/ruff/rules/#pydoclint-doc
```
damskii9992: You also accidentally reverted this to comment out the DOC rule :)

damskii9992 (commented):
I think we should have an ADR on the logger to explain why we implemented it, and instructions on how to use it, both when writing to it, and when ignoring it (I guess the last one can simply refer to the .md file you wrote).

docs/docs/user-guide/controlling-log-output.md
rozyczko: `autouse` fixture `yield`s *inside* `with at_level(logging.ERROR)`, so the test body runs within the context and **is** muted.  Your `set_level('ERROR')` also works and is marginally simpler, but the code is correct

docs/docs/user-guide/controlling-log-output.md
rozyczko: Python's `logging.lastResort` handler emits WARNINGs and above to stderr when no handlers are
configured, so standalone `easyscience` users *do* see warnings/errors out of the box.
Only INFO/DEBUG are hidden without change of setup.

src/easyscience/utils/decorators.py
rozyczko: `stacklevel` *is* a supported kwarg on `logging` methods and here it correctly points the warning at the caller of the deprecated function. 
Not having it elsewhere is a fair point. I'll add it to other deprecation log locations for consistency

src/easyscience/global_object/logger.py
rozyczko: silently falling back to a default on a bad value is the _right_ behaviour. You don't want  import to crash on a typo'd env var. This is for env var parsing.
FOr public API (set_level/at_level) tthis should raise. I think we need to split the concerns here 
Strict parse for the API
lenient parse (default) for env var

src/easyscience/base_classes/collection_base.py
rozyczko: I don't think this is an issue.
Moving from `warnings.warn(..., DeprecationWarning)` to logging indeed removes the warning filtering by category, `-W` flags, pytest `filterwarnings`). 
But - categories are replaced by the**logger-name hierarchy** (`easyscience.legacy`, `easyscience.deprecated`,  easyscience.fitting.bumps`, …), which is filterable per-subsystem.

pyproject.toml
rozyczko: Cause pixi run check is messing it up!!!
I need to change it once and for all

rozyczko:
> I think we should have an ADR on the logger to explain why we implemented it, and instructions on how to use it, both when writing to it, and when ignoring it (I guess the last one can simply refer to the .md file you wrote).

https://github.com/easyscience/core/discussions/277

src/easyscience/global_object/logger.py
```diff
@@ -1,44 +1,189 @@
...
+    'ERROR': logging.ERROR,
+    'CRITICAL': logging.CRITICAL,
+}
+
+
+def _parse_log_level(level: int | str) -> int:
+    """Convert a level number or name into a numeric level, raising on an unknown name."""
+    if isinstance(level, int):
+        return level
+    stripped = level.strip()
```
damskii9992: What if level was something else than an int or str?
Since you're using the method in the level setter, you need to check the input type.

damskii9992 (approved):
LGTM now with a single addition ;)

damskii9992:
And don't forget to finalize the ADR suggestion into a real ADR:
https://github.com/easyscience/core/discussions/277
And maybe link to the discussions in this PR?

src/easyscience/global_object/logger.py
rozyczko: Good catch. this is called directly from the setter, so isn't guaranteed the right type.

PR #257, opened 2026-06-04
Bayesian extend/resume

rozyczko:
Support for resuming and extending Bayesian MCMC chains using the BUMPS DREAM sampler.
Added ability to save and reload sampler state and continue chains across sessions. 

* **Resume and extend MCMC chains:**
  - Added `resume_state` parameter to `mcmc_sample` and `sample` methods in both `fitter.py` and `minimizer_bumps.py`, allowing users to continue sampling from a saved chain state. This includes documentation of the "ring-buffer contract" for extending chains without losing samples.

* **Population/chains argument resolution:**
  - Introduced `_resolve_population_alias` method to handle both `chains` and `population` arguments, raising an error if both are set with different values, improving API clarity and preventing silent bugs.


src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -263,6 +264,41 @@ def _resolve_fitclass(method: str):
...
+        ------
+        ValueError
+            If both ``chains`` and ``population`` are provided but differ.
+        """
+        if chains is not None and population is not None and chains != population:
+            raise ValueError(
+                f'Conflicting population specifications: chains={chains} '
+                f'and population={population}. Use only one.'
+            )
+        return chains if chains is not None else population
```
damskii9992: Why did you re-introduce this?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -415,8 +453,40 @@ def mcmc_sample(
...
+            ``mcmc_sample()`` call (e.g. ``PosteriorResults.sampler_state``).
+            When provided, DREAM **continues** the saved chain instead of
+            starting cold.  The population, parameter count, and parameter
+            names must match the current model — a ``ValueError`` is raised
+            otherwise.
+
+            **Ring-buffer contract (important!):** DREAM stores draws in a
+            fixed-size ring buffer sized to *samples*.  Resuming with
+            ``samples=N`` retains only the **last N** draws.  To extend an
+            existing chain of M draws by N without losing any::
```
damskii9992: All you really need to write here is: `samples` must be the total number of samples, so to extend by `M`, `samples` should be `N + M`.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -415,8 +453,40 @@ def mcmc_sample(
...
+
+                fitter.mcmc_sample(data, samples=M + N, burn=0, resume_state=old_state)
+
+            The ``burn`` parameter controls burn-in for the *new* draws
+            only; passing ``burn=0`` (strongly recommended on resume)
+            skips additional burn-in.  A non-zero ``burn`` on a
+            previously-converged chain is usually a mistake and emits a
+            warning.
```
damskii9992: Should we maybe force it to be 0 when resuming?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -415,8 +453,40 @@ def mcmc_sample(
...
         thin : int, default=10
             Thinning interval.
+        chains : int | None, default=None
+            User-friendly alias for ``population``.  Provide one or the
+            other, not both.
         population : int | None, default=None
-            BUMPS DREAM population count (number of parallel chains).
+            DREAM population **scale factor** (not an absolute chain count):
+            BUMPS creates ``ceil(population * n_parameters)`` parallel chains,
+            so the default scale of 10 yields ``10 * n_parameters`` chains.
```
damskii9992: Maybe just write:
"BUMPS DREAM population count per parameter (number of parallel chains).

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -415,8 +453,40 @@ def mcmc_sample(
...
+
+            The ``burn`` parameter controls burn-in for the *new* draws
+            only; passing ``burn=0`` (strongly recommended on resume)
+            skips additional burn-in.  A non-zero ``burn`` on a
+            previously-converged chain is usually a mistake and emits a
+            warning.
+
+            The ``chains``/``population`` and ``initializer`` parameters
+            have **no effect** when ``resume_state`` is provided — they
+            are determined by the saved state.
```
damskii9992: I think it might be a cleaner API if we make the resume a new method without these ignored parameters, and then `resume_state` would be obligatory.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -479,10 +552,102 @@ def mcmc_sample(
...
+            )
+            if labels_carry_our_names:
+                state_stripped = [lbl[len(state_prefix) :] for lbl in state_labels]
+                if fresh_names != state_stripped:
+                    raise ValueError(
+                        f'Parameter names/order mismatch between the current model '
+                        f'and resume_state.\n'
+                        f'  Current model : {fresh_names}\n'
+                        f'  resume_state  : {state_stripped}'
+                    )
```
damskii9992: I feel like we should save the state on the BUMPS object, so that when resuming in a notebook, the user doesn't have to supply the `resume_state` to resume. They would only need to supply it when resuming from a file. That would also clean up this code, as you know that when this keyword is supplied, the state is always resumed from a file. And I think it would provide a nicer user-experience.

damskii9992 (commented):
Just gonna stop my review here as there is already plenty to work on.
As we discussed in-person, I also think we should refactor the bayesian sampling to a seperate "samplers" module. This should help us when we add other Bayesian samplers in the future as well as provide a cleaner user API (the user doesn't create a "Fitter" when doing sampling).
It also more cleanly separates responsibility in the code. A `Minimizer` minimizes a function to the data, a `Sampler` samples the parameters to create distributions.
To avoid duplicating code, it might be smart to move shared functionality to new base-classes.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: We might need to. I'll change that.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -250,20 +250,20 @@
...
     "\n",
-    "DREAM only works with the BUMPS minimizer. We reuse the ``mle_fitter`` created above, switch it to BUMPS, and call `Fitter.mcmc_sample`, which returns a dictionary with the following keys:\n",
+    "DREAM only works with the BUMPS minimizer. We reuse the ``mle_fitter`` created above, switch it to BUMPS, and create a `Sampler` instance bound to the fitter and data. Calling `sampler.sample()` returns a `SamplingResults` object with the following attributes:\n",
     "\n",
     "- `draws`: a `(n_samples, n_parameters)` array of posterior samples: each **row** is one complete draw from the joint posterior (one value for every parameter simultaneously), and each **column** holds all sampled values for a single parameter;\n",
     "- `param_names`: the unique names of the parameters, in the same column order as `draws`;\n",
-    "- `internal_bumps_object`: the underlying BUMPS `MCMCDraw` object, useful for advanced diagnostics;\n",
+    "- `state`: the underlying BUMPS `MCMCDraw` object, useful for advanced diagnostics;\n",
```
damskii9992, line 268: I assume reverting this attribute name is a mistake?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -492,11 +485,125 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "03339658",
+   "metadata": {},
+   "source": [
+    "## Extend the chain and check convergence\n",
+    "\n",
+    "The original run used ``samples=10000`` (5000 retained after thinning).  Here we **extend the chain by 5000 more raw samples** using ``sampler.extend()`` — DREAM continues from the previous in-memory state instead of starting cold.\n",
```
damskii9992: Would probably write "instead of starting from scratch" rather than "starting from cold"

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -492,11 +485,125 @@
...
+    "\n",
+    "``sampler.extend(additional_samples=5000)`` does the ring-buffer arithmetic for you automatically: it reads the number of stored generations from the saved state and sizes the new buffer to ``old_generations + additional_samples`` so no existing draws are lost.\n",
+    "\n",
+    "After the extension we compare the posterior summaries and check convergence with Gelman-Rubin R-hat."
```
damskii9992: What is Gelman-Rubin R-hat? Could you explain it a bit?

pixi.lock
damskii9992, line 1: There is a bug in the current scipy/bumps versions. You need to run another pixi update. Otherwise your tutorial notebook doesn't run.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -492,11 +485,125 @@
...
+   "source": [
+    "# Extend the existing chain by 5000 more raw samples.\n",
+    "# extend() handles the ring-buffer arithmetic automatically.\n",
+    "extended_results = sampler.extend(additional_samples=5000, thin=2)\n",
+    "\n",
+    "short_draws = results.draws\n",
+    "extended_draws = extended_results.draws\n",
+    "\n",
+    "print(f'Short chain: {short_draws.shape[0]} retained draws')\n",
+    "print(f'Extended chain: {extended_draws.shape[0]} retained draws')"
```
damskii9992: The output here is that the extended chain has 7000 draws. Adding 5000 with a thin of 2 should make the chain 2500 longer. Clearly we're applying the burn-in of 500 draws again, but why are we doing that when we extend the chain? Shouldn't that only happen when you initiate the chain?

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -492,11 +485,125 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Convergence diagnostic: Gelman-Rubin R-hat from the extended chain state.\n",
+    "# Values close to 1.0 indicate good convergence.\n",
+    "print('Gelman-Rubin R-hat (extended chain) — values < 1.05 indicate convergence:')\n",
+    "rhat = extended_results.state.gelman()\n",
+    "for name, val in zip(extended_results.param_names, rhat):\n",
+    "    status = '✓' if val < 1.05 else '?' if val < 1.1 else '✗'\n",
+    "    print(f'  {name:<20s} {val:.4f}  {status}')"
```
damskii9992: Shouldn't you compare these values with the ones from the 5000 draws?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+_SIDECAR_SCHEMA_VERSION = 2
+_ACCEPTED_SIDECAR_SCHEMA_VERSIONS = (1, 2)
+
+
+def _easyscience_version() -> str:
+    """Return the installed easyscience version string."""
+    try:
+        from importlib.metadata import version as _v
+
+        return _v('easyscience')
```
damskii9992: why `as _v`? Also, shouldn't this method be in the utils?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+    w_list: list,
+) -> Optional[str]:
+    """Return a SHA-256 hex digest of concatenated (x|y|weights), or None."""
+    try:
+        h = hashlib.sha256()
+        for arr in list(x_list) + list(y_list) + list(w_list):
+            h.update(np.ascontiguousarray(arr, dtype=np.float64).tobytes())
+        return h.hexdigest()
+    except Exception:
+        return None
```
damskii9992, line 48: Why the `return None` branch? Should it raise an error?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+def save_chain(
+    state: Any,
```
damskii9992: Didn't we talk about not calling this the "state", but rather "Bumps_dream_object" or something?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+    param_names: Optional[List[str]],
```
damskii9992: We don't use `Optional` for type hinting, use `list[str] | None`.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+from .minimizers.minimizer_base import MINIMIZER_PARAMETER_PREFIX
+
+if TYPE_CHECKING:  # avoid import cycles; Fitter is only needed for type hints
+    from .fitter import Fitter
+
+# Sidecar schema owned by easyscience from version 2 onwards. Version 1
+# sidecars (written by easyreflectometry's save_posterior) carry only
+# 'param_names' and are still accepted on load.
+_SIDECAR_SCHEMA_VERSION = 2
+_ACCEPTED_SIDECAR_SCHEMA_VERSIONS = (1, 2)
```
damskii9992: We should DEFINITELY have a discussion and an ADR on loading/saving and backwards compatibility.

damskii9992: Also, what even is a sidecar?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+    try:
+        h = hashlib.sha256()
+        for arr in list(x_list) + list(y_list) + list(w_list):
+            h.update(np.ascontiguousarray(arr, dtype=np.float64).tobytes())
+        return h.hexdigest()
+    except Exception:
+        return None
+
+
+def save_chain(
```
damskii9992: Shouldn't this be a method on the `Sampler` object?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+    state: Any,
+    param_names: Optional[List[str]],
+    path: str,
+    data_fingerprint: Optional[str] = None,
+) -> None:
+    """Persist a DREAM chain state plus a metadata sidecar.
+
+    Writes BUMPS native files (``<path>-chain.mc.gz`` etc. via
+    ``bumps.dream.state.save_state``) plus a ``<path>.params.json`` sidecar
+    holding parameter names and metadata. Used by :meth:`Sampler.save` and by
```
damskii9992: This was written by an AI I assume, we don't use the notation :meth: that is sphinx notation.

src/easyscience/fitting/sampler.py
damskii9992: Also, why is this typehinted as `Any`? Shouldn't it be the DREAMS objects?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+        Parameter names (one per chain column), stored in the sidecar.
+    path : str
+        File path prefix. BUMPS appends its own suffixes.
+    data_fingerprint : Optional[str], default=None
+        SHA-256 fingerprint of the data the chain was sampled against,
+        verified (with a warning) on reload.
+    """
+    from .minimizers.minimizer_bumps import save_sampler_state
+
+    save_sampler_state(state, str(path))
```
damskii9992: ??? You have another save function elsewhere which you import and call here?
Why not consolidate them into a single function?

src/easyscience/fitting/sampler.py
damskii9992: Also, doesn't this need a default to actually be optional?

damskii9992: And finally, optional arguments come after mandatory arguments.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+        # save_state/load_state does not preserve labels, so a reloaded state
+        # typically carries default labels like ['P0', 'P1', ...].
+        param_names = [
+            lbl[len(MINIMIZER_PARAMETER_PREFIX) :]
+            if lbl.startswith(MINIMIZER_PARAMETER_PREFIX)
+            else lbl
+            for lbl in state.labels
+        ]
+
+    return state, param_names, sidecar
```
damskii9992: Many of the same comments here from the save function also apply to the load function.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+    }
+    with open(f'{path}.params.json', 'w') as f:
+        json.dump(sidecar, f, indent=2)
+
+
+def load_chain(path: str, skip: int = 0) -> tuple[Any, Optional[List[str]], dict]:
+    """Reload a DREAM chain state saved by :func:`save_chain`.
+
+    Uses the patched loader (``load_sampler_state``), which carries the
+    BUMPS >= 1.0.4 single-row ``loadtxt`` workaround. Parameter names are
```
damskii9992: This looks like AI again. The docstring shouldn't tell you that the internal function it uses has a workaround . . .
And again, :meth:, :func: etc are sphinx notation.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+    restored from the sidecar when available (schema versions 1 and 2),
+    falling back to the state's labels with the minimizer prefix stripped.
+    Used by :meth:`Sampler.load_state` and by the deprecated
+    ``easyreflectometry`` ``load_posterior`` shim.
```
damskii9992: Probably also shouldn't write where a function is used in the functions docstring . . .

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+        'schema_version': _SIDECAR_SCHEMA_VERSION,
+        'param_names': param_names,
+        'easyscience_version': _easyscience_version(),
+        'data_fingerprint': data_fingerprint,
+    }
+    with open(f'{path}.params.json', 'w') as f:
+        json.dump(sidecar, f, indent=2)
+
+
+def load_chain(path: str, skip: int = 0) -> tuple[Any, Optional[List[str]], dict]:
```
damskii9992: How can the output be typehinted as Optional? Huh?

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+
+    sidecar: dict = {}
+    param_names: Optional[List[str]] = None
+    try:
+        with open(f'{path}.params.json', 'r') as f:
+            sidecar = json.load(f)
+        if sidecar.get('schema_version') in _ACCEPTED_SIDECAR_SCHEMA_VERSIONS:
+            param_names = sidecar.get('param_names')
+    except (FileNotFoundError, json.JSONDecodeError):
+        sidecar = {}
```
damskii9992, line 214: codecov says this misses a test.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+class Sampler:
+    """Bayesian MCMC sampler for one dataset, backed by a Fitter's BUMPS minimizer.
+
+    One ``Sampler`` instance represents one chain over one ``(x, y, weights)``
+    dataset. The data is bound at construction; :meth:`sample` and
+    :meth:`extend` take no data arguments, so a chain can never be extended
+    against different data (undefined behaviour in BUMPS).
+
+    Construct directly with a configured ``Fitter`` (or ``MultiFitter``) whose
+    minimizer has been switched to ``AvailableMinimizers.Bumps``.
```
damskii9992: When the `Sampler` is constructed from the `Fitter` it appears like you have to do a fit first, which you don't.
Instead of having the `Sampler` take a `Fitter` as input, I would make the `Sampler` an independent class.
Maybe it can have a constructor `Sampler.from_fitter` which **Constructs** a `Sampler` from a `Fitter` object, but I wouldn't make it required.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,541 @@
...
+
+    The sampler is BUMPS/DREAM-specific for now: the BUMPS check in
+    :meth:`_run` is the seam where another backend would plug in.
+
+    Parameters
+    ----------
+    fitter : Fitter
+        A configured ``Fitter`` (or ``MultiFitter``) whose minimizer has been
+        switched to ``AvailableMinimizers.Bumps``.
+    x : Union[np.ndarray, List[np.ndarray]]
```
damskii9992: . . . Is this a Union typehint in the docstring? XD
Wow, the autocomplete is strong in this one.

src/easyscience/fitting/sampler.py
damskii9992: Also, we should probably copy over the structure of the minimizers. I.e. with a factory, a list of available samplers (for now just DREAM) and separate backend classes.

damskii9992 (commented):
I'm gonna stop my review again here as the changes I have proposed are again substantial and another review would likely be required afterwards.

My biggest issue lies with the class/file structure. I think we should copy the factory/minimizer structure for the samplers, as this is an easily extendable and logically sounds structure and it makes reviews easier as the structure would be the same across our codebase. But we probably need an ADR for this and to have discussion about it, probably in-person.

rozyczko:
> My biggest issue lies with the class/file structure. I think we should copy the factory/minimizer structure for the samplers, as this is an easily extendable and logically sounds structure and it makes reviews easier as the structure would be the same across our codebase. But we probably need an ADR for this and to have discussion about it, probably in-person.

There is exactly one sampler (DREAM) and one backend.
Building a factory + enum + base class for a single implementation is, although potentially useful, slightly overengineering at this point. The `Sampler._run` BUMPS check is already documented as the place for a second backend. I would suggest creating an ADR, but not modifying this PR. Let's do it when another sampler actually gets implemented, so the abstraction is derived from two real cases rather than guessed from one.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Added a paragraph.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: If we ever want to add another sampler, this would have to be changed. It _is_ a sampler state, so calling it such is not confusing.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Nope. the burn-in is not re-applied. `Sampler.extend()` passes burn=0, and `Bumps.mcmc_sample` sets burn=0 whenever resume_state is not None, warning if the caller asked otherwise. There is no path that re-burns a resumed chain.

rozyczko: I'll update the notebook/docs with proper statements

src/easyscience/fitting/sampler.py
rozyczko: `easyscience/__init__.py` already does __version__ = version('easyscience').
_easyscience_version() is redundant. Deleted

src/easyscience/fitting/sampler.py
rozyczko: It's deliberate. The fingerprint is best-effort attempt, and an un-fingerprintable array must NOT block a save that took hours of sampling.

src/easyscience/fitting/sampler.py
rozyczko: As discussed above.

src/easyscience/fitting/sampler.py
rozyczko: the only defaulted argument is last

src/easyscience/fitting/sampler.py
rozyczko: 1. 100% agree on the ADR. 
2. A sidecar is a small JSON file written next to the BUMPS chain files (``<path>.params.json``). The BUMPS chain format doesn't store our parameter names nor any ownership info, so we keep them alongside it. Added a description

src/easyscience/fitting/sampler.py
rozyczko: `save_chain/load_chain` are module-level only to satisfy easyreflectometry's `save_posterior/load_posterior`. I guess it's better to just put it on Sampler and modify the library.

src/easyscience/fitting/sampler.py
rozyczko: fixed

src/easyscience/fitting/sampler.py
rozyczko: irrelevant, removed the method

src/easyscience/fitting/sampler.py
rozyczko: fixed

src/easyscience/fitting/sampler.py
rozyczko: types are HARD, ok??

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: 2-parameter model, `sample(samples=10000, burn=500, thin=2)` then `extend(additional_samples=5000, thin=2))`:

| Config | `sample()` draws | `extend()` draws | Delta |
| --- | --- | --- | --- |
| default (`trim=True`) | 4520 | 6600 | +2080 |
| `sampler_kwargs={'trim': False}` | **5000** | **7500** | **+2500** |

With trimming off, the numbers are *exactly* as you expect. The ring buffer also grows exactly as intended: `Nthin * Npop == 7500`, and `state.draw(portion=1.0, outliers=True)` returns all 7500 points. **No draws are lost, and none are re-burned.**

Running the actual tutorial (4 parameters, not the 2 above) reproduces your number exactly and shows where it comes from:

| chain | `results.draws` (trimmed) | full chain |
| --- | --- | --- |
| original | 4400 | 5000 |
| extended | **7000** | **7600** |

So the "missing" draws were never missing: the chain grew from 5000 to 7600.

src/easyscience/fitting/sampler.py
rozyczko: Modified.

src/easyscience/fitting/sampler.py
rozyczko: Think of the cost: `Sampler` reaches into `Fitter` internals (_precompute_reshaping, _fit_function_wrapper, and assigning fitter.fit_function to swap the wrapper.
Making it standalone means extracting that machinery into a shared base, which will affect the basic functionality of the fitter. I woudn't want to do it unless really, really needed.

I will reword the docstring/notebook to say no prior fit is needed as a solution.

rozyczko:
> > My biggest issue lies with the class/file structure. I think we should copy the factory/minimizer structure for the samplers, as this is an easily extendable and logically sounds structure and it makes reviews easier as the structure would be the same across our codebase. But we probably need an ADR for this and to have discussion about it, probably in-person.
> 
> There is exactly one sampler (DREAM) and one backend. Building a factory + enum + base class for a single implementation is, although potentially useful, slightly overengineering at this point. The `Sampler._run` BUMPS check is already documented as the place for a second backend. I would suggest creating an ADR, but not modifying this PR. Let's do it when another sampler actually gets implemented, so the abstraction is derived from two real cases rather than guessed from one.

https://github.com/easyscience/core/discussions/280

src/easyscience/fitting/sampler.py
rozyczko: https://github.com/orgs/easyscience/discussions/72

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -492,11 +496,200 @@
...
+    "\n",
+    "$$\n",
+    "\\hat{R} \\approx \\sqrt{\\frac{\\text{between-chain variance} + \\text{within-chain variance}}{\\text{within-chain variance}}}\n",
+    "$$\n",
+    "\n",
+    "The intuition is that of a spot-the-difference test. If every chain has converged to\n",
+    "the same distribution, then the spread *across* chains is just the spread *within* one\n",
+    "chain seen from another angle, both variance estimates agree, and $\\hat{R} \\to 1$. If\n",
+    "the chains are still drifting toward the posterior, or have settled in different\n",
+    "regions of it, the between-chain variance is inflated and $\\hat{R}$ rises above 1.\n",
```
damskii9992: This doesn't make sense. If `between-chain variance` approaches the `within-chain variance`, then according to the equation "R" should go towards sqrt(2), not 1. It can only go towards 1 if `between-chain variance` goes towards zero, or becomes MUCH smaller than `within-chain variance`.

docs/docs/tutorials/fitting-bayesian.ipynb
```diff
@@ -492,11 +496,200 @@
...
+    "may place the burn point differently each time. It can even shrink.\n",
+    "\n",
+    "This is *not* the `burn=500` being applied a second time — `extend()` always resumes\n",
+    "with `burn=0`, and BUMPS forces it to 0 on resume regardless.\n",
+    "\n",
+    "Nothing is lost: the trimming is only a view on the ring buffer. Read the full chain\n",
+    "with `results.state.draw(portion=1.0, outliers=True)`, or switch trimming off with\n",
+    "`sampler_kwargs={'trim': False}`. Either way you get the whole chain, as the next cell\n",
+    "demonstrates. (Even then, don't expect `samples / thin` on the nose: DREAM allocates\n",
+    "its buffer in whole generations, so `samples` is a floor that gets rounded up.)\n",
```
damskii9992: This section sounds SO AI-like, and unfortunately I think that makes it appear like we don't know what is going on and just had an AI write it. . .

docs/docs/tutorials/fitting-bayesian.ipynb
damskii9992: Okay, but this still doesn't explain everything.
So the "Full" chain included the burn-in but not the thinned-out, since the first full chain is 5000 steps, but the extended is 2600 larger, which is more than half of the 5000 asked for. You write that DREAM rounds up to nearest whole generation, so I guess that must be 200 steps? Is that the default or is that a parameter we can change? It is weird that it runs longer than you ask it to.

src/easyscience/fitting/sampler.py
damskii9992: You should AT LEAST report a warning then, maybe using your fancy-schmancy new logging module? The user should be notified that the saved state lost its identification fingerprint.

src/easyscience/fitting/sampler.py
damskii9992: So you don't import a save function now, but you still have both a function and a method on the class, why not consolidate them? They're quite small functions/methods both.

src/easyscience/fitting/sampler.py
damskii9992: I understand why you're reluctant, this is the easy solution, but it is also the messy solution which will compound our technical debt. By not doing this now, you're just postponing this task and making it harder in the future. It also means that WHEN we do this task in the future, it is gonna be an API-breaking change, whereas by adding it correctly from the get-go isn't API-breaking.

rozyczko:
I added the plan for refactoring in https://github.com/easyscience/core/discussions/280

This follows discussion in Slack.

I would prefer addressing this in a separate PR.

src/easyscience/fitting/sampler.py
damskii9992: Following our discussion on Slack, I will reluctantly relent this issue to a (soon) future PR.

damskii9992 (commented):
Answers to comments, will continue with my review.

src/easyscience/fitting/minimizers/__init__.py
```diff
@@ -8,4 +8,11 @@
 from .utils import FitError
 from .utils import FitResults
 
-__all__ = [MinimizerBase, Bumps, DFO, LMFit, FitError, FitResults]
```
damskii9992, line 11: I recently learned that all the `__all__` attributed does, is change what gets imported when you do star imports, i.e.:
```python
from easyscience.fitters.minimizers import *
```
The default for that is to import all modules, so simply adding all modules to an `__all__` is completely redundant . . .

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -389,6 +390,7 @@ def mcmc_sample(
         burn: int = 2000,
         thin: int = 10,
         population: int | None = None,
+        resume_state: Any | None = None,
```
damskii9992: I assume this shouldn't be typehinted `Any`?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -482,10 +509,100 @@ def mcmc_sample(
...
+                        f'Requested population ({pop}) would produce '
+                        f'{expected_npop} chains but the saved state has '
+                        f'{resume_state.Npop} chains. The population cannot '
+                        f'be changed on resume.'
+                    )
+            else:
+                # Recover the original scale factor from the state's Npop.
+                # Try ceil-division first, then fall back to floor.
+                recovered = int(math.ceil(resume_state.Npop / n_params))
+                if int(math.ceil(recovered * n_params)) != resume_state.Npop:
```
damskii9992: You KNOW that `recovered` and `n_params` are integers, hence you don't need the int and math.ceil calls here.
Also, what are you even trying to do here? You get the floor rounding always, UNLESS `resume_state.Npop / n_params` is an integer-like float, in which case the ceil function does the same as int?
This section kinda seems redundant.

src/easyscience/fitting/minimizers/minimizer_bumps.py
damskii9992, line 388: This method is getting stupidly long now with the resume functionality, please try to factor out some functionality into helper methods to ease readability.

src/easyscience/fitting/__init__.py
```diff
@@ -4,8 +4,10 @@
...
 from .fitter import Fitter
 from .minimizers.utils import FitResults
+from .sampler import Sampler
+from .sampler import SamplingResults
 
 # Causes circular import
 # from .multi_fitter import MultiFitter  # noqa: F401, E402
 
-all = [AvailableMinimizers, Fitter, FitResults]
+all = [AvailableMinimizers, Fitter, FitResults, Sampler, SamplingResults]
```
damskii9992: As per my other comment.

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Eww. Bad take on R̂...
Fixed.
B is defined as the variance of the per-chain means, with the correct convergence argument (each chain mean has standard error sqrt(W/n), so B ≈ W/n → 0 and R̂ → 1)

docs/docs/tutorials/fitting-bayesian.ipynb
rozyczko: Meh, reworded it to be more consistent. By hand.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+    def _loadtxt_2d(file, report: int = 0) -> np.ndarray:
+        arr = loader(file, report=report)
+        # A single saved row is read back as 1-D; restore it to a (1, N) row.
+        return np.atleast_2d(arr) if getattr(arr, 'ndim', 2) == 1 else arr
+
+    bumps_state.loadtxt_with_fallback = _loadtxt_2d
+    try:
+        return bumps_state.load_state(path, skip=skip)
+    finally:
+        bumps_state.loadtxt_with_fallback = loader
```
damskii9992: Why bother writing code to support bumps versions older than 1.0.4?
1.0.4 is already previous release, just require bumps >= 1.0.4 in pyproject.toml and remove the other branch.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+        File path prefix. BUMPS appends its own suffixes.
+    data_fingerprint : str | None, default=None
+        SHA-256 fingerprint of the data the chain was sampled against,
+        verified (with a warning) on reload.
+    """
+    from bumps.dream.state import save_state
+
+    from easyscience import __version__
+
+    save_state(state, str(path))
```
damskii9992: If this is a user-function, you should check your input, i.e. that state is a MDMDDRaw object, that path is a str etc. If it isn't a user-function, it should start with an _.

src/easyscience/fitting/sampler.py
rozyczko: OK, Added the check/warning in the callers, since the message depends on whether it's state load or save.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+        ``bumps.dream.state.load_state``.
+
+    Returns
+    -------
+    tuple[MCMCDraw, list[str] | None, dict]
+        The reloaded BUMPS chain state, the parameter names (or ``None`` if
+        neither sidecar nor labels yielded them), and the raw sidecar dict
+        (empty if absent/unreadable).
+    """
+    state = _load_bumps_state(str(path), skip=skip)
```
damskii9992: Same comment as for the save function.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+        sampler_kwargs: dict | None = None,
+    ):
+        self._fitter = fitter
+        self._x = x
+        self._y = y
+        self._weights = weights
+        self._vectorized = vectorized
+        self._default_sampler_kwargs = dict(sampler_kwargs or {})
+        self._state: MCMCDraw | None = None  # current chain state
+        self._results: SamplingResults | None = None
```
damskii9992, line 414: No checks on any of the inputs?

rozyczko:
Issues from the most recent round addressed.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+
+    @property
+    def param_names(self) -> list[str] | None:
+        """Parameter names from the most recent run (or None)."""
+        return self._results.param_names if self._results is not None else None
+
+    @property
+    def logp(self) -> np.ndarray | None:
+        """Log-posterior values from the most recent run (or None)."""
+        return self._results.logp if self._results is not None else None
```
damskii9992, line 459: No properties for the x, y, weights or fitter, etc?
Mind, if you write them their setters should either be disabled or create a new internal bumps object, to avoid having a mismatch.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+        x: np.ndarray | list[np.ndarray],
+        y: np.ndarray | list[np.ndarray],
+        weights: np.ndarray | list[np.ndarray | None] | None = None,
+        vectorized: bool = False,
+        sampler_kwargs: dict | None = None,
+    ):
+        self._fitter = fitter
+        self._x = x
+        self._y = y
+        self._weights = weights
```
damskii9992: If you set these as copies of the inputs, then you avoid the issue that the user might change these after having them set on the object. Which I think is reasonable.

src/easyscience/fitting/sampler.py
damskii9992: If you choose to make the x, y, weights and fitter copies of the given inputs, you should have setters for their properties, so the user can in fact change them after creating the object (with a reset of the bumps internal object).

src/easyscience/fitting/sampler.py
damskii9992, line 1: There is currently a lot of BUMPS-specific code in this file, which makes sense since that is the only sampler we currently have, but bear in mind that WHEN we have another sampler, there should be NO bumps-specific code or syntax in this file, those should all live in the minimizer_bumps.py file.
All syntax here should be sampler-independent.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+        population: int | None = None,
+        sampler_kwargs: dict | None = None,
+        progress_callback: Callable[[dict], bool | None] | None = None,
+        abort_test: Callable[[], bool] | None = None,
+    ) -> SamplingResults:
+        """Run fresh Bayesian MCMC sampling on the bound data.
+
+        Calling ``sample()`` on a sampler that already holds a chain starts a
+        **fresh** chain — the previous state and results are replaced. Use
+        ``extend()`` to continue an existing chain.
```
damskii9992, line 552: Maybe add a warning if a user tries to run sample() when there already is a result, something like:
"Warning, replacing previous chain with fresh chain, if you meant to continue the chain, please use the extend() method"
Or something along those lines.

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,651 @@
...
+        saved_fingerprint = sidecar.get('data_fingerprint')
+        if saved_fingerprint is not None:
+            current_fingerprint = self._fingerprint()
+            if current_fingerprint is not None and current_fingerprint != saved_fingerprint:
+                warnings.warn(
+                    'The data bound to this Sampler does not match the data '
+                    'fingerprint stored with the saved chain. Extending a '
+                    'chain against different data is undefined behaviour.',
+                    UserWarning,
+                    stacklevel=2,
```
damskii9992: You should probably use your own logging system here.

tests/integration/fitting/test_multi_fitter.py
```diff
@@ -288,213 +288,5 @@ def test_multi_fit_1D_2D(fit_engine):
...
-            samples=100,
-            burn=20,
-            thin=2,
-            sampler_kwargs={'init': 'random'},
-        )
-
-        assert result['draws'].ndim == 2
-        assert result['draws'].shape[0] > 0
+# NOTE: The Bayesian MCMC tests formerly here (TestMultiFitterMcmcSample) were
+# moved to test_sampler.py and adapted to ``Sampler(f, ...)``.
```
damskii9992: We probably don't need this comment lol.

tests/integration/fitting/test_sampler.py
damskii9992, line 1: Most of these tests belong in unit tests. Remember, the unit tests folder should mimic the src folder. So sampler.py should have a mirror test_samlper.py in unit tests.

Integration tests are only for full runs, like running an actual sampling and checking the resulting R value or convergence or the value of some parameter or some such. 
It would also be the place to have a full save/load/resume roundtrip test.

damskii9992 (changes requested):
I did not look closely at the tests, neither did I pay too much attention to the structure since we plan to change that, but here are the last comments I found after going through the PR again.

src/easyscience/fitting/minimizers/__init__.py
rozyczko: `__all__` isn't just defaults. It declares the public API and can control  *import of names. But those shouldn't really be objects, just strings. Anyway, removed since we are now exposing all the methods anyway.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Modified to ceil-division (and a comment)

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: resume and validation can be a separate method indeed. Refactored

src/easyscience/fitting/sampler.py
rozyczko: removed the cruft. Updated the toml file

src/easyscience/fitting/sampler.py
rozyczko: added exhaustive checks.

src/easyscience/fitting/sampler.py
rozyczko: > with a reset of the bumps internal object

Nope. a chain should never be extended against different data. Settable `x` allows this travesty. "Create a new Sampler for new data" is a better solution, I think

src/easyscience/fitting/sampler.py
rozyczko: makes sense to make defensive copies here.

src/easyscience/fitting/sampler.py
rozyczko: deferred to the planned refactor.

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -380,6 +385,99 @@ def _make_func(x, y, weights):
...
+        if burn > 0:
+            logger.warning(
+                f'burn={burn} ignored on resume: a previously converged '
+                f'chain is not re-burned. Forcing burn=0.'
+            )
+
+        # A negative ``pop`` is read by ``bumps.initpop.generate`` as an
+        # absolute chain count, exactly reproducing the saved population
+        # without having to recover the original scale factor.
+        return -int(resume_state.Npop), 0
```
damskii9992: You're not dividing Npop by n_params anymore, is that not a mistake?

src/easyscience/fitting/minimizers/minimizer_bumps.py
```diff
@@ -380,6 +385,99 @@ def _make_func(x, y, weights):
 
         return _outer(self)
 
+    def _validate_resume_state(
```
damskii9992: I personally prefer having helper methods appear after the methods they're used in, but maybe that's just my preference :)

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,751 @@
...
+    not path-like and ``ValueError`` if ``skip`` is not a non-negative int.
+    """
+    if not isinstance(path, (str, os.PathLike)):
+        raise TypeError(f'path must be a str or os.PathLike, got {type(path).__name__}.')
+    if isinstance(skip, bool) or not isinstance(skip, int) or skip < 0:
+        raise ValueError('skip must be a non-negative integer.')
+    return str(path)
+
+
+def _load_bumps_state(path: str, skip: int = 0) -> MCMCDraw:
```
damskii9992, line 133: codecov is complaining that this function does not have tests

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,751 @@
...
+        x: np.ndarray | list[np.ndarray],
+        y: np.ndarray | list[np.ndarray],
+        weights: np.ndarray | list[np.ndarray | None] | None = None,
+        vectorized: bool = False,
+        sampler_kwargs: dict | None = None,
+    ):
+        if not (hasattr(fitter, 'minimizer') and hasattr(fitter, 'fit_function')):
+            raise TypeError(
+                f'fitter must be a configured Fitter or MultiFitter, got {type(fitter).__name__}.'
+            )
```
damskii9992, line 375: I believe this check is redundant. I don't think we are even capable of creating a Fitter without these attributes. At least not with the public API

src/easyscience/fitting/sampler.py
```diff
@@ -0,0 +1,751 @@
...
+        x_is_multi = isinstance(x, (list, tuple))
+        if x_is_multi != isinstance(y, (list, tuple)):
+            raise ValueError('x and y must either both be arrays or both be lists of arrays.')
+        if x_is_multi and len(x) != len(y):
+            raise ValueError(
+                f'x and y must hold the same number of datasets, got {len(x)} and {len(y)}.'
```
damskii9992, line 381: Your checks have holes, what if x or y are scalars? or strings? Please make proper input checks.

damskii9992 (changes requested):
It looks like a decent chunk of your code still needs unit tests, according to all the codecov warnings I am seeing. Otherwise, there's only a few more things.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: Nope. Returning `-Npop` reproduces the saved population exactly.
If you look at the BUMPS source: bumps/initpop.py:54 does 
`pop_size = ceil(pop * len(initial)) if pop > 0 else int(-pop)`

A negative pop is BUMPS' documented convention for an absolute chain count.

src/easyscience/fitting/sampler.py
rozyczko: The check isn't just checking the attribute, it guards against passing the wrong object entirely (a model, a minimizer, None) to a public constructor.
Ohterwise, we would get a random AttributeError somehere deep in _run() rather than immendiate TypeError here.

src/easyscience/fitting/minimizers/minimizer_bumps.py
rozyczko: low cost fix and I don't object

damskii9992 (approved):
Lets get this merged in.

PR #282, opened 2026-07-27
added fix for #281

### LINKED ISSUE #281: Dependent Parameters do not update the calculator
When a dependent `Parameter` is bound to a calculator through `_callback` (in `InterfaceFactoryTemplate.generate_bindings`), changes to
its independent component never propagate to the calculator, and the recomputed value is overwritten with the old calculator value on the next `.value` read.

The result: **every constraint with a bound dependent parameter is completely neglected during fitting and simulation.**

In easyreflectometry this breaks:

- `MaterialMixture`: fitting `fraction` does not change its value, and `fraction` has zero effect on the computed reflectivity (`_sld`/`_isld` are dependent parameters bound to the calculator),
- gradient-layer thickness/roughness constraints (each sub-layer's thickness/roughness is dependent  on the assembly parameter),
- surfactant/bilayer area-per-molecule, solvation and roughness constraints.

No error or warning is raised anywhere.

rozyczko:
Dependent parameters did not properly push their updated values to a callback when their dependencies changed. 

This fixes it by calling the callback set function with the updated value any time the value changes.

### DIFF
--- a/src/easyscience/variable/parameter.py
+++ b/src/easyscience/variable/parameter.py
@@ -235,6 +235,9 @@ class Parameter(DescriptorNumber):
             if self._desired_unit is not None:
                 self._convert_unit(self._desired_unit)
 
+            if self._callback.fset is not None:
+                self._callback.fset(self._scalar.value)
+
             self._notify_observers()
         else:
             global_object.log.getLogger('variable').warning(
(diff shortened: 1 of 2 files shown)


damskii9992 (approved):
LGTM

Although I find it really stupid that we keep 2 copies of each paramters value stored. One in the `Parameter` and one in the `interface`. We should just remove the one in the `interface` and have the interface get its value from the `Parameter` instead of having this stupid syncronization, but that is a task for later :/

